PR #1, opened 2025-07-17
[spring-ecommerce-product - Step 1-1 to 1-3] Submit Deniz' mission

deniz-oezdemir:
Hey there,

Hope you are doing great! 👋 

We just wrapped up steps 1-1 to 1-3 and had a lot of fun diving into Spring and web development for the first time. We used the learning tasks as a starting point and built our own version of the API from there.

- `Product` is our model
- `ProductController` handles the CRUD operations
- `ProductRepository` handles the SQL statements

A couple of things we’d love your input on:
- We ran into some test issues with the database - specifically, the tests in `SSRTest` only pass when run by themselves, but fail when all tests are run together. It has something to do with test interference (as the tests in different test files mock the database in different ways) but we could not figure out how to fix that.
- Since we followed the learning tasks pretty closely, we are curious if there is a better or more typical way to structure things in Spring.

We are both brand new to Spring, so any feedback or advice would be super appreciated.

This implementation is ready for review, thanks so much for taking the time to look it over! 🙏 

All the best,
Deniz

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,5 +11,5 @@
 - [ ] UI representation of products
   - [x] retrieve
   - [x] add
-  - [ ] update
+  - [x] update
   - [x] delete
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 - [ ] configure API that services the following requests:
   - [x] add
-  - [ ] retrieve
+  - [x] retrieve
   - [ ] update
   - [ ] delete
 
--- a/README.md
+++ b/README.md
@@ -2,9 +2,8 @@
 
 ## feature list
 
-- [ ] configure API that services the following requests:
+- [x] configure API that services the following requests:
   - [x] add
   - [x] retrieve
   - [x] update
   - [x] delete
-
--- a/README.md
+++ b/README.md
@@ -1,1 +1,10 @@
 # spring-ecommerce-product
+
+## feature list
+
+- [ ] configure API that services the following requests:
+  - [x] add
+  - [ ] retrieve
+  - [ ] update
+  - [ ] delete
+
--- a/README.md
+++ b/README.md
@@ -8,8 +8,10 @@
   - [x] update
   - [x] delete
 
-- [ ] UI representation of products
+- [x] UI representation of products
   - [x] retrieve
   - [x] add
   - [x] update
   - [x] delete
+
+- [x] database integration (instead of global properties)
--- a/README.md
+++ b/README.md
@@ -7,3 +7,5 @@
   - [x] retrieve
   - [x] update
   - [x] delete
+
+- [ ] UI representation of products
--- a/README.md
+++ b/README.md
@@ -9,3 +9,7 @@
   - [x] delete
 
 - [ ] UI representation of products
+  - [x] retrieve
+  - [x] add
+  - [ ] update
+  - [x] delete
--- a/README.md
+++ b/README.md
@@ -6,5 +6,5 @@
   - [x] add
   - [x] retrieve
   - [x] update
-  - [ ] delete
+  - [x] delete
 
--- a/README.md
+++ b/README.md
@@ -5,6 +5,6 @@
 - [ ] configure API that services the following requests:
   - [x] add
   - [x] retrieve
-  - [ ] update
+  - [x] update
   - [ ] delete
 
--- a/src/main/kotlin/ecommerce/Product.kt
+++ b/src/main/kotlin/ecommerce/Product.kt
@@ -1,0 +1,17 @@
+package ecommerce
+
+class Product(
+    var id: Long? = null,
+    var name: String,
+    var price: Double,
+    var imageUrl: String,
+) {
+    companion object {
+        fun toEntity(
+            product: Product,
+            id: Long,
+        ): Product {
+            return Product(id, product.name, product.price, product.imageUrl)
+        }
+    }
+}
--- a/src/main/kotlin/ecommerce/Product.kt
+++ b/src/main/kotlin/ecommerce/Product.kt
@@ -6,12 +6,18 @@ class Product(
     var price: Double,
     var imageUrl: String,
 ) {
+    fun update(newProduct: Product) {
+        this.name = newProduct.name
+        this.price = newProduct.price
+        this.imageUrl = newProduct.imageUrl
+    }
+
     companion object {
         fun toEntity(
             product: Product,
             id: Long,
         ): Product {
-            return Product(id, product.name, product.price, product.imageUrl)
+            return Product(id = id, product.name, product.price, product.imageUrl)
         }
     }
 }
--- a/src/main/kotlin/ecommerce/Product.kt
+++ b/src/main/kotlin/ecommerce/Product.kt
@@ -6,18 +6,21 @@ class Product(
     var price: Double,
     var imageUrl: String,
 ) {
-    fun update(newProduct: Product) {
-        this.name = newProduct.name
-        this.price = newProduct.price
-        this.imageUrl = newProduct.imageUrl
+    init {
+        require(name.length <= 255)  {RuntimeException("[Error] Name can be maximum 255 characters long.")}
+        require(imageUrl.length <= 512) {RuntimeException("[Error] URL can be maximum 512 characters long.")}
     }
 
-    companion object {
-        fun toEntity(
-            product: Product,
-            id: Long,
-        ): Product {
-            return Product(id = id, product.name, product.price, product.imageUrl)
-        }
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as Product
+
+        return id == other.id
+    }
+
+    override fun hashCode(): Int {
+        return id?.hashCode() ?: 0
     }
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -14,7 +14,6 @@ import java.net.URI
 
 @Controller
 class ProductController(private val productRepository: ProductRepository) {
-
     @PostMapping("/api/products")
     fun createProduct(
         @RequestBody product: Product,
@@ -46,7 +45,7 @@ class ProductController(private val productRepository: ProductRepository) {
         @PathVariable id: Long,
     ): ResponseEntity<Void> {
         val deleted = productRepository.delete(id)
-        if (!deleted){
+        if (!deleted) {
             throw RuntimeException()
         }
         return ResponseEntity.noContent().build()
@@ -60,7 +59,7 @@ class ProductController(private val productRepository: ProductRepository) {
 }
 
 @Controller
-class ProductPageController (private val productRepository: ProductRepository) {
+class ProductPageController(private val productRepository: ProductRepository) {
     @GetMapping("/products")
     fun getProducts(model: Model): String {
         model.addAttribute("products", productRepository.getAll())
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -2,6 +2,7 @@ package ecommerce
 
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Controller
+import org.springframework.ui.Model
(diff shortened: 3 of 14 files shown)


README.md
```diff
@@ -1 +1,17 @@
...
+  - [x] update
+  - [x] delete
+
+- [x] UI representation of products
+  - [x] retrieve
+  - [x] add
+  - [x] update
+  - [x] delete
+
+- [x] database integration (instead of global properties)
```
Hyeon9mak, line 17: 😉 👍 nice habbit

src/main/kotlin/ecommerce/Product.kt
```diff
@@ -0,0 +1,23 @@
...
+    fun update(newProduct: Product) {
+        this.name = newProduct.name
+        this.price = newProduct.price
+        this.imageUrl = newProduct.imageUrl
+    }
+
+    companion object {
+        fun toEntity(
+            product: Product,
+            id: Long,
```
Hyeon9mak: In the context of entities, equality is determined based on the ID.

For example, if you have

- id = 1, name = "red apple"

On the other hand

```kotlin

        other as Product

        return id == other.id
    }

        return id?.hashCode() ?: 0
    }
```

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import java.net.URI
+
+@Controller
+class ProductController(private val productRepository: ProductRepository) {
```
Hyeon9mak: What's different between `@RestController` & `@Controller` ?

Hyeon9mak: and u can also append this annotation 😉 
what is it's role?

```kotlin
@RequestMapping("/api/products")
```

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+        @RequestBody newProduct: Product,
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        val updated = productRepository.update(id, newProduct)
+        if (!updated) {
+            throw RuntimeException("Product not found")
+        }
+        return ResponseEntity.ok().build()
+    }
+
```
Hyeon9mak: nice choice POST, GET, PUT, DELETE methods! 💯 

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+    @DeleteMapping("/api/products/{id}")
+    fun deleteProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        val deleted = productRepository.delete(id)
+        if (!deleted) {
+            throw RuntimeException()
+        }
+        return ResponseEntity.noContent().build()
```
Hyeon9mak, line 54: I think delete checking is not presentation(controller) layer's responsibility. 🤔 
how do u think? what is presentation layer's role?

src/main/resources/schema.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE products (
+                          id BIGINT AUTO_INCREMENT PRIMARY KEY,
+                          name VARCHAR(255) NOT NULL,
+                          price DOUBLE NOT NULL,
+                          image_url VARCHAR(512) NOT NULL
+);
```
Hyeon9mak: name & image_url has limits!

I think adding validation logic to the `Product`, might make the code more robust and easier to test.
U think so too, right? 😄

src/main/resources/static/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <title>Products</title>
+</head>
+<body>
+Hello, Friend!
+</body>
```
Hyeon9mak, line 9: Hi~

Hyeon9mak (changes requested):
Hi Deniz! Welcome to the Spring mission. 😄 

You’ve implemented everything very cleanly!

~~When running all the tests together, the error below occurs.~~
~~If you have time, it would be great to try fixing it.~~

image

This happens because table creation in each test overlaps with the one defined in schema.sql.
One way to fix it would be to create a separate configuration file that doesn’t rely on schema.sql.

`spring.sql.init.mode=never`

Once you resolve this issue, you’ll likely encounter a port conflict between tests.
To fix that, try using RANDOM_PORT instead of DEFINED_PORT. 😉 

image

README.md
deniz-oezdemir: Thanks!

src/main/kotlin/ecommerce/Product.kt
deniz-oezdemir: Just to be sure, are we adding `equals()` and `hashCode()` now purely for future-proofing?
At the moment, we’re not comparing Product instances directly (we rely on the database to match by id during updates), so I want to confirm that this is a preventive change for possible future use in collections, caching, or ORM behavior.

Pleas let me know if that's the case here! ☺️

src/main/kotlin/ecommerce/ProductController.kt
deniz-oezdemir: Thanks for the suggestion.

`@Controller` is used for traditional MVC controllers that return views (like HTML via Thymeleaf), while `@RestController` is used for APIs that return JSON.

We updated the controller to use `@RestController` instead of `@Controller`. This fits better since the class returns JSON, not views.

We also added `@RequestMapping("/api/products")` at the class level. This let us simplify the method-level mappings.

src/main/kotlin/ecommerce/ProductController.kt
deniz-oezdemir: PATCH would allow to update product properties partially.

PUT requires all product properties to be updated.

For now we believe PUT is sufficient (also because our UI requires input for each property). But please let us know if we need to implement PATCH. 😉

src/main/resources/schema.sql
deniz-oezdemir: Yes, I added `require` statements in the `Product` initiator. Now, in case the user inputs are not valid, they do not even reach the database. Also I added tests to verify the behavior.

src/main/resources/static/index.html
deniz-oezdemir: 👋

src/main/kotlin/ecommerce/ProductController.kt
deniz-oezdemir: That's a very good question! 🤔 

You're right that the presentation layer shouldn't handle deciding whether a delete is valid. It should only handle HTTP request processing.

The check for whether the product exists (and the logic for what to do if it doesn't) fits better in a service layer.

If we had a service layer, that would be a good place for this logic. Since we’re using `productRepository` directly here, we could either:
- Move the existence check and error handling into the `productRepository`, or
- introduce a service layer to manage that responsibility more clearly.

As for now we would prefer to leave it as is since there are only two checks in `updateProduct` and `deleteProduct`. When we continue with the next step and build on the project, would it be better to add a service layer, or keep it simple and push this logic into the `productRepository`?

deniz-oezdemir:
Hi @Hyeon9mak,

Thanks for your review. I made the updates based on your feedback.

I left two issues unresolved since I had a couple of questions there and would really appreciate your input. Whenever you get a chance, could you take another look? 

Let me know if there’s anything else I should tweak for this step. 😊

Best, Deniz

src/main/kotlin/ecommerce/Product.kt
```diff
@@ -0,0 +1,26 @@
...
+
+class Product(
+    var id: Long? = null,
+    var name: String,
+    var price: Double,
+    var imageUrl: String,
+) {
+    init {
+        require(name.length <= 255)  {RuntimeException("[Error] Name can be maximum 255 characters long.")}
+        require(imageUrl.length <= 512) {RuntimeException("[Error] URL can be maximum 512 characters long.")}
```
Hyeon9mak, line 11: 😉 👍

src/main/kotlin/ecommerce/Product.kt
```diff
@@ -0,0 +1,26 @@
...
+        if (javaClass != other?.javaClass) return false
+
+        other as Product
+
+        return id == other.id
+    }
+
+    override fun hashCode(): Int {
+        return id?.hashCode() ?: 0
+    }
```
Hyeon9mak, line 25: > Just to be sure, are we adding equals() and hashCode() now purely for future-proofing?

That's right — at the moment, it doesn't serve any functional purpose.
But it's helpful to think of it as something you've prepared in advance for your teammates.

Code is not only a way to communicate with the compiler, but also a form of shared documentation between developers.
By adding this, you're subtly telling your teammates, "This class is meant to be used as an entity."
And since we can’t always predict how other developers will use the code later, it's a small step to help them understand the intention behind it.

In a way, it’s about preparing for the “what ifs” and making the code more maintainable and intuitive for everyone. 😄

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,71 @@
...
+    @PostMapping
+    fun createProduct(
+        @RequestBody product: Product,
+    ): ResponseEntity<Void> {
+        val id = productRepository.create(product)
+        return ResponseEntity.created(URI.create("/api/products/$id")).build()
+    }
+
+    @GetMapping
+    fun readProducts(): ResponseEntity<List<Product>> {
```
Hyeon9mak, line 29: 😉 👍

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,71 @@
...
+    fun updateProduct(
+        @RequestBody newProduct: Product,
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        val updated = productRepository.update(id, newProduct)
+        if (!updated) {
+            throw RuntimeException("Product not found")
+        }
+        return ResponseEntity.ok().build()
+    }
```
Hyeon9mak, line 44: > For now we believe PUT is sufficient (also because our UI requires input for each property). But please let us know if we need to implement PATCH. 😉

I expected u to understand PATCH vs PUT in detail and make a thoughtful choice.
And that’s exactly what you did. Well done! 👍 💯

src/main/kotlin/ecommerce/ProductController.kt
Hyeon9mak: That’s a perfect thinking and response. 💯 
Since you’re already aware of the issue, I think it’s totally fine to move forward as it is at this stage 😄

src/test/kotlin/ecommerce/ProductTest.kt
```diff
@@ -0,0 +1,48 @@
...
+    @Test
+    fun `should throw when name exceeds 255 characters`() {
+        val longName = "a".repeat(256)
+
+        assertThrows(RuntimeException::class.java) {
+            Product(
+                name = longName,
+                price = 2.49,
+                imageUrl = "https://example.com/pistachio.jpg"
+            )
```
Hyeon9mak, line 31: niiiiiiiice 💯

Hyeon9mak (approved):
Hi Deniz, So sorry about late review 🙏 

All requirements have been met, and the tests are passing nicely!
The level of thought you’re putting into this is already quite advanced — great work 💯 

The parts you were considering can definitely be improved in the next step 😉 
Let’s move on to the next stage!

src/main/kotlin/ecommerce/Product.kt
deniz-oezdemir: Thanks for clarifying! 😃

src/main/kotlin/ecommerce/ProductController.kt
deniz-oezdemir: Perfect, thank you!

PR #2, opened 2025-07-17
[spring-ecommerce-product - Step 1-1 to 1-3] Submit Mariia' mission

mariyakulikova:
Hey there,
Hope you are doing great! :wave:
We just wrapped up steps 1-1 to 1-3 and had a lot of fun diving into Spring and web development for the first time. We used the learning tasks as a starting point and built our own version of the API from there.
- `Product` is our model
- `ProductController` handles the CRUD operations
- `ProductRepository` handles the SQL statements

A couple of things we’d love your input on:
- We ran into some test issues with the database - specifically, the tests in `SSRTest` only pass when run by themselves, but fail when all tests are run together. It has something to do with test interference (as the tests in different test files mock the database in different ways) but we could not figure out how to fix that.
- Since we followed the learning tasks pretty closely, we are curious if there is a better or more typical way to structure things in Spring.

We are both brand new to Spring, so any feedback or advice would be super appreciated.
This implementation is ready for review, thanks so much for taking the time to look it over! :pray:

All the best,
Mariia

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -29,6 +29,7 @@ dependencies {
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("io.rest-assured:rest-assured:5.3.1")
 }
 
 kotlin {
--- a/src/main/kotlin/ecommerce/Product.kt
+++ b/src/main/kotlin/ecommerce/Product.kt
@@ -1,0 +1,17 @@
+package ecommerce
+
+class Product(
+    var id: Long? = null,
+    var name: String,
+    var price: Double,
+    var imageUrl: String,
+) {
+    companion object {
+        fun toEntity(
+            product: Product,
+            id: Long,
+        ): Product {
+            return Product(id, product.name, product.price, product.imageUrl)
+        }
+    }
+}
--- a/src/main/kotlin/ecommerce/Product.kt
+++ b/src/main/kotlin/ecommerce/Product.kt
@@ -6,12 +6,18 @@ class Product(
     var price: Double,
     var imageUrl: String,
 ) {
+    fun update(newProduct: Product) {
+        this.name = newProduct.name
+        this.price = newProduct.price
+        this.imageUrl = newProduct.imageUrl
+    }
+
     companion object {
         fun toEntity(
             product: Product,
             id: Long,
         ): Product {
-            return Product(id, product.name, product.price, product.imageUrl)
+            return Product(id = id, product.name, product.price, product.imageUrl)
         }
     }
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -10,11 +10,14 @@ import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
 import java.net.URI
 
-@Controller
+@RestController
+@RequestMapping("api/products")
 class ProductController(private val productRepository: ProductRepository) {
-    @PostMapping("/api/products")
+    @PostMapping
     fun createProduct(
         @RequestBody product: Product,
     ): ResponseEntity<Unit> {
@@ -22,13 +25,13 @@ class ProductController(private val productRepository: ProductRepository) {
         return ResponseEntity.created(URI.create("/api/products/$id")).build()
     }
 
-    @GetMapping("/api/products")
+    @GetMapping
     fun readProducts(): ResponseEntity<List<Product>> {
         val products = productRepository.getAll()
         return ResponseEntity.ok(products)
     }
 
-    @PutMapping("/api/products/{id}")
+    @PutMapping("/{id}")
     fun updateProduct(
         @RequestBody newProduct: Product,
         @PathVariable id: Long,
@@ -40,7 +43,7 @@ class ProductController(private val productRepository: ProductRepository) {
         return ResponseEntity.ok().build()
     }
 
-    @DeleteMapping("/api/products/{id}")
+    @DeleteMapping("/{id}")
     fun deleteProduct(
         @PathVariable id: Long,
     ): ResponseEntity<Unit> {
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -17,7 +17,7 @@ class ProductController(private val productRepository: ProductRepository) {
     @PostMapping("/api/products")
     fun createProduct(
         @RequestBody product: Product,
-    ): ResponseEntity<Void> {
+    ): ResponseEntity<Unit> {
         val id = productRepository.create(product)
         return ResponseEntity.created(URI.create("/api/products/$id")).build()
     }
@@ -32,7 +32,7 @@ class ProductController(private val productRepository: ProductRepository) {
     fun updateProduct(
         @RequestBody newProduct: Product,
         @PathVariable id: Long,
-    ): ResponseEntity<Void> {
+    ): ResponseEntity<Unit> {
         val updated = productRepository.update(id, newProduct)
         if (!updated) {
             throw RuntimeException("Product not found")
@@ -43,7 +43,7 @@ class ProductController(private val productRepository: ProductRepository) {
     @DeleteMapping("/api/products/{id}")
     fun deleteProduct(
         @PathVariable id: Long,
-    ): ResponseEntity<Void> {
+    ): ResponseEntity<Unit> {
         val deleted = productRepository.delete(id)
         if (!deleted) {
             throw RuntimeException()
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -14,7 +14,6 @@ import java.net.URI
 
 @Controller
 class ProductController(private val productRepository: ProductRepository) {
-
     @PostMapping("/api/products")
     fun createProduct(
         @RequestBody product: Product,
@@ -46,7 +45,7 @@ class ProductController(private val productRepository: ProductRepository) {
         @PathVariable id: Long,
     ): ResponseEntity<Void> {
         val deleted = productRepository.delete(id)
-        if (!deleted){
+        if (!deleted) {
             throw RuntimeException()
         }
         return ResponseEntity.noContent().build()
@@ -60,7 +59,7 @@ class ProductController(private val productRepository: ProductRepository) {
 }
 
 @Controller
-class ProductPageController (private val productRepository: ProductRepository) {
+class ProductPageController(private val productRepository: ProductRepository) {
     @GetMapping("/products")
     fun getProducts(model: Model): String {
         model.addAttribute("products", productRepository.getAll())
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -2,6 +2,7 @@ package ecommerce
 
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Controller
+import org.springframework.ui.Model
 import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.GetMapping
@@ -12,10 +13,26 @@ import org.springframework.web.bind.annotation.RequestBody
 import java.net.URI
 import java.util.concurrent.atomic.AtomicLong
 
+// TODO: remove mock data for product list
+//val MOCK_PRODUCTS = listOf(
+//    Product(name = "melon ice cream", price = 2.33, imageUrl = "link-1.jpg"),
+//    Product(name = "lemon ice cream", price = 1.50, imageUrl = "link-2.jpg"),
+//    Product(name = "vanilla ice cream", price = 3.09, imageUrl = "link-3.jpg")
+//)
+//
+//val products: MutableMap<Long, Product> = MOCK_PRODUCTS
+//    .map { product ->
+//        val id = index.getAndIncrement()
+//        id to Product.toEntity(product, id)
+//    }
+//    .toMap()
+//    .toMutableMap()
+
+val products: MutableMap<Long, Product> = HashMap()
+val index = AtomicLong(3)
+
 @Controller
 class ProductController {
-    private val products: MutableMap<Long, Product> = HashMap()
-    private val index = AtomicLong(1)
 
     @PostMapping("/api/products")
     fun createProduct(
@@ -63,5 +80,14 @@ class ProductController {
     @ExceptionHandler(RuntimeException::class)
     fun handle(e: RuntimeException): ResponseEntity<Unit> {
         return ResponseEntity.notFound().build()
+    }
+}
+
+@Controller
+class ProductPageController {
+    @GetMapping("/products")
+    fun getProducts(model: Model): String {
+        model.addAttribute("products", products.values.toList())
+        return "products"
     }
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -3,6 +3,7 @@ package ecommerce
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Controller
 import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
@@ -56,5 +57,11 @@ class ProductController {
         if (deletedProduct == null) throw RuntimeException()
 
         return ResponseEntity.noContent().build()
+    }
+
+    // Intercepts any thrown RuntimeException within ProductController
+    @ExceptionHandler(RuntimeException::class)
+    fun handle(e: RuntimeException): ResponseEntity<Unit> {
+        return ResponseEntity.notFound().build()
     }
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -90,6 +90,4 @@ class ProductPageController {
         model.addAttribute("products", products.values.toList())
         return "products"
     }
-
-
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -1,0 +1,24 @@
+package ecommerce
+
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import java.net.URI
+import java.util.concurrent.atomic.AtomicLong
+
+@Controller
+class ProductController {
+    private val products: MutableMap<Long, Product> = HashMap()
+    private val index = AtomicLong(1)
+
+    @PostMapping("/api/products")
+    fun createProduct(
+        @RequestBody product: Product,
+    ): ResponseEntity<Void> {
+        val id = index.getAndIncrement()
+        val newProduct: Product = Product.toEntity(product, id)
+        products.put(id, newProduct)
+        return ResponseEntity.created(URI.create("api/products/$id")).build()
+    }
+}
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -11,42 +11,22 @@ import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import java.net.URI
-import java.util.concurrent.atomic.AtomicLong
-
-// TODO: remove mock data for product list
-//val MOCK_PRODUCTS = listOf(
-//    Product(name = "melon ice cream", price = 2.33, imageUrl = "link-1.jpg"),
-//    Product(name = "lemon ice cream", price = 1.50, imageUrl = "link-2.jpg"),
-//    Product(name = "vanilla ice cream", price = 3.09, imageUrl = "link-3.jpg")
-//)
-//
-//val products: MutableMap<Long, Product> = MOCK_PRODUCTS
-//    .map { product ->
-//        val id = index.getAndIncrement()
-//        id to Product.toEntity(product, id)
-//    }
-//    .toMap()
-//    .toMutableMap()
-
-val products: MutableMap<Long, Product> = HashMap()
-val index = AtomicLong(1)
 
 @Controller
-class ProductController {
+class ProductController(private val productRepository: ProductRepository) {
 
     @PostMapping("/api/products")
     fun createProduct(
         @RequestBody product: Product,
     ): ResponseEntity<Void> {
-        val id = index.getAndIncrement()
-        val newProduct: Product = Product.toEntity(product, id)
-        products.put(id, newProduct)
-        return ResponseEntity.created(URI.create("api/products/$id")).build()
+        val id = productRepository.create(product)
+        return ResponseEntity.created(URI.create("/api/products/$id")).build()
     }
 
     @GetMapping("/api/products")
     fun readProducts(): ResponseEntity<List<Product>> {
-        return ResponseEntity.ok(products.values.toList())
+        val products = productRepository.getAll()
+        return ResponseEntity.ok(products)
     }
 
     @PutMapping("/api/products/{id}")
@@ -54,14 +34,10 @@ class ProductController {
         @RequestBody newProduct: Product,
         @PathVariable id: Long,
     ): ResponseEntity<Void> {
-        val currentProduct = products[id]
-
-        if (currentProduct == null) {
-            throw RuntimeException()
+        val updated = productRepository.update(id, newProduct)
+        if (!updated) {
+            throw RuntimeException("Product not found")
         }
-
-        currentProduct.update(newProduct)
-
         return ResponseEntity.ok().build()
     }
 
@@ -69,10 +45,10 @@ class ProductController {
     fun deleteProduct(
         @PathVariable id: Long,
     ): ResponseEntity<Void> {
-        val deletedProduct = products.remove(id)
-
-        if (deletedProduct == null) throw RuntimeException()
-
+        val deleted = productRepository.delete(id)
+        if (!deleted){
+            throw RuntimeException()
+        }
         return ResponseEntity.noContent().build()
     }
 
@@ -84,10 +60,10 @@ class ProductController {
 }
 
 @Controller
-class ProductPageController {
+class ProductPageController (private val productRepository: ProductRepository) {
     @GetMapping("/products")
     fun getProducts(model: Model): String {
-        model.addAttribute("products", products.values.toList())
+        model.addAttribute("products", productRepository.getAll())
         return "products"
     }
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -2,6 +2,7 @@ package ecommerce
 
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Controller
+import org.springframework.web.bind.annotation.DeleteMapping
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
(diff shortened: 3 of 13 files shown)


build.gradle.kts
```diff
@@ -29,6 +29,7 @@ dependencies {
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("io.rest-assured:rest-assured:5.3.1")
```
pkeugine, line 32: In our previous missions, we didn't use this library when making tests.
What do you think this library brings to us when testing our application?
I ask this because I think knowing our tools and why we need it is as important as knowing how to use it :)

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+    @PostMapping("/api/products")
+    fun createProduct(
+        @RequestBody product: Product,
+    ): ResponseEntity<Void> {
+        val id = productRepository.create(product)
+        return ResponseEntity.created(URI.create("/api/products/$id")).build()
+    }
+
+    @GetMapping("/api/products")
+    fun readProducts(): ResponseEntity<List<Product>> {
```
pkeugine, line 29: I see you've used `ResponseEntity` as a return type in `controller`.
In our previous missions, we would call `OutputView` since our objective was to print the info in the console. What is the role of this `ResponseEntity`? Are there any other options we can choose from?

src/main/kotlin/ecommerce/Product.kt
```diff
@@ -0,0 +1,23 @@
...
+        this.imageUrl = newProduct.imageUrl
+    }
+
+    companion object {
+        fun toEntity(
+            product: Product,
+            id: Long,
+        ): Product {
+            return Product(id = id, product.name, product.price, product.imageUrl)
+        }
```
pkeugine, line 21: What do you think this term`entity` means?
Also, when we want to set an id to a `product` we can simply:
```kotlin
        val product = Product(
            id = id,
            name = "product",
            price = 99.99,
            imageUrl = "https://example.com/sample.png",
        )
```
Why would we want to use this `toEntity` method instead, and should we enforce it? 🤔

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import java.net.URI
+
+@Controller
```
pkeugine: I see you've used `@Controller` annotation!
There's also `@RestController` annotation, so let's compare these annotations and see which one fits our needs. It would be great if you could tell me what you've researched 👍

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+class ProductController(private val productRepository: ProductRepository) {
+    @PostMapping("/api/products")
```
pkeugine: I see there's `/api` prefix in front of some URLs.
Can you tell me why they were added, and why only some of them have it?

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+    fun createProduct(
+        @RequestBody product: Product,
+    ): ResponseEntity<Void> {
+        val id = productRepository.create(product)
+        return ResponseEntity.created(URI.create("/api/products/$id")).build()
```
pkeugine, line 25: `ResponseEntity.created`, `ResponseEntity.ok`, `ResponseEntity.notFound`, `ResponseEntity.noContent`... So many different phrases related to http status!

What do they mean, and why did you use these in each cases?
I guess their names kind of tell us what they mean, but it'll be nice to actually look up detailed explanations and tell me about it :)

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+        if (!updated) {
+            throw RuntimeException("Product not found")
+        }
+        return ResponseEntity.ok().build()
+    }
+
+    @DeleteMapping("/api/products/{id}")
+    fun deleteProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
```
pkeugine:
```suggestion
    ): ResponseEntity<Unit> {
```

This is not a must, but we can actually use `Unit` here as well.
What are the differences?
+) after looking up the differences, what would you use?

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+        val deleted = productRepository.delete(id)
+        if (!deleted) {
+            throw RuntimeException()
+        }
+        return ResponseEntity.noContent().build()
+    }
+
+    // Intercepts any thrown RuntimeException within ProductController
```
pkeugine: It would be great to remove this comment.
Although I do understand this comment was created in the process of learning,
it doesn't seem necessary to put it in our final branch.

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+    @ExceptionHandler(RuntimeException::class)
+    fun handle(e: RuntimeException): ResponseEntity<Unit> {
+        return ResponseEntity.notFound().build()
+    }
+}
+
+@Controller
+class ProductPageController(private val productRepository: ProductRepository) {
+    @GetMapping("/products")
+    fun getProducts(model: Model): String {
```
pkeugine, line 66: `@RequestBody newProduct: Product`
`@PathVariable id: Long`
`model: Model`

Unlike the previous missions, there are so many different kinds of parameters!
What are their differences, and are there any options we could use when using annotations such as `@ResponseBody` and `@PathVariable`?

**Hint**: looking at their implementation might help:
```kotlin
public @interface RequestBody {
    boolean required() default true;
}
```

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import java.net.URI
+
+@Controller
+class ProductController(private val productRepository: ProductRepository) {
```
pkeugine, line 19: Where do you think this `productRepository` object is coming from? 
I don't see any logic that creates this object. 🤔
Maybe it's getting created in the background with the help of `@Repository` annotation in `ProductRepository` class?

src/test/kotlin/ecommerce/CRUDTest.kt
```diff
@@ -0,0 +1,160 @@
...
+
+    @BeforeEach
+    fun setUp() {
+        productRepository = ProductRepository(jdbcTemplate)
+
+        jdbcTemplate.execute("DROP TABLE products IF EXISTS")
+        jdbcTemplate.execute(
+            "CREATE TABLE products(" +
+                "id BIGINT AUTO_INCREMENT, name VARCHAR(255) NOT NULL, price DOUBLE NOT NULL, image_url VARCHAR(512) NOT NULL)",
+        )
```
pkeugine, line 29: Why should we drop the table and create the same one **`BeforeEach` tests**?
I mean, we might as well reuse the one that we created before and save some time 🤔 

Oh, I'm not suggesting removal. I actually think it's great.
I'm asking this because I used to throw this question at myself when creating tests. 🙂

src/test/kotlin/ecommerce/CRUDTest.kt
```diff
@@ -0,0 +1,160 @@
...
+                        name = "Orange ice cream",
+                        price = 2.80,
+                        imageUrl = "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcRf8Fdb-33SCOszWX_UF-92pCwX4Rcam0uVCg&s",
+                    ),
+                )
+                .contentType(ContentType.JSON)
+                .`when`().post("/api/products")
+                .then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.CREATED.value())
```
pkeugine:
```suggestion
                .`when`().post("/api/products")
                .then().log().all()
                .assertThat().statusCode(HttpStatus.CREATED.value())
```

When you have the time, I recommend reading the official documentation of the libraries that we use. `RestAssured` itself offers a lot of asserting tools!

This is just a comment to show that there are so many cool things we don't know about the things that we use, but in practice, there are often times when we add bunch of dependencies only to find out that it could've been done with only one : _the one we already had for several years_ 😂

src/test/kotlin/ecommerce/CRUDTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    @Test
+    fun delete_failure() {
+        val response =
+            RestAssured
+                .given().log().all()
+                .`when`().delete("/api/products/4")
+                .then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.NOT_FOUND.value())
+    }
```
pkeugine: Nice to see testing failing scenario 💯

src/test/kotlin/ecommerce/CRUDTest.kt
```diff
@@ -0,0 +1,160 @@
...
+    fun read() {
+        val response =
+            RestAssured
+                .given().log().all()
+                .contentType(ContentType.JSON)
+                .`when`().get("/api/products")
+                .then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.OK.value())
+        assertThat(response.jsonPath().getList("", Product::class.java)).hasSize(3)
```
pkeugine: What if we changed the `@BeforeEach` logic and added more test fixtures?
I think the number `3` is vulnerable to this kind of change. 🤔

src/test/kotlin/ecommerce/DBTest.kt
```diff
@@ -0,0 +1,98 @@
+package ecommerce
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.JdbcTest
+import org.springframework.jdbc.core.JdbcTemplate
+
+@JdbcTest
```
pkeugine, line 10: I see this `@JdbcTest` annotation is a different choice from the previous
```kotlin
@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
```
which were in `CRUDTest`. What does `@JdbcTest` do?

src/test/kotlin/ecommerce/DBTest.kt
```diff
@@ -0,0 +1,98 @@
...
+class DBTest {
+    private lateinit var productRepository: ProductRepository
+
+    @Autowired
+    private lateinit var jdbcTemplate: JdbcTemplate
```
pkeugine, line 15: Yet another annotation!
What does `@Autowired` annotation do?

src/test/kotlin/ecommerce/DBTest.kt
```diff
@@ -0,0 +1,98 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.JdbcTest
+import org.springframework.jdbc.core.JdbcTemplate
+
+@JdbcTest
+class DBTest {
+    private lateinit var productRepository: ProductRepository
```
pkeugine, line 12: Why should we `lateinit` this `ProductRepository`?

src/test/kotlin/ecommerce/SSRTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.test.annotation.DirtiesContext
+
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@AutoConfigureMockMvc
```
pkeugine: Yet another annotation! (part 2)
What does `@AutoConfigureMockMvc` annotation do?

src/test/kotlin/ecommerce/SSRTest.kt
```diff
@@ -0,0 +1,88 @@
...
+    @Test
+    fun `HTML page does not contain product name`() {
+        val response =
+            RestAssured
+                .get("/products")
+
+        val html = response.body.asString()
+
+        assertThat(html).doesNotContain("melon")
+    }
```
pkeugine: I'm not 100% convinced about why we would want to test if the page contains "melon" or not.
I'm guessing this test is part of the template, but it would be nice to change the test name or the assertion so that anyone can understand the purpose of the test!

src/test/kotlin/ecommerce/SSRTest.kt
```diff
@@ -0,0 +1,88 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.test.annotation.DirtiesContext
+
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@AutoConfigureMockMvc
+class SSRTest() {
```
pkeugine, line 14: **Q. We ran into some test issues with the database - specifically, the tests in SSRTest**

I would like you to read the logs after running all the tests altogether.
I know the log seems very dramatic(?) with all the red sentences and it being like 10000 lines long,
but if we read it, we are able to see the 'main reason' why SSLTests are failing.
Try focusing on the trailing `Caused by` logs such as:
Screenshot 2025-07-19 at 2 44 21 AM

I think this would be a great experience tracking down the problem using logs;
I do it all the time, and so will you (it's not always fun but... yeah...)

After solving the issue in the log, I was able to make all tests pass :  

Screenshot 2025-07-19 at 2 41 31 AM

Oh, and there are multiple ways to solve current problem, so don't try too much on finding the best solution.

**Hint**: it's related to database table.

pkeugine (changes requested):
Hello Mariia, and welcome to the world of Spring ... Boot  🍀 😎 
You've done a great job!

Since it's our first time using it, I focused on `how`s and `why`s rather than `better`s and `improvements`. That resulted in asking bunch of questions 😅 but I'm 100% sure these questions will help you understand more about our programming environment.

I strongly recommend answering all the questions, even if it takes some time. If it's too hard to answer, hit me with comments or DMs and we'll tackle them together. (But I suggest tackling the test issue on your own. I think it's an opportunity for a great skill gain.)

If you're done, send me another review request right away!
Happy Hacking! 🍀 🍀 🍀

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: Since we're building a web API, ResponseEntity plays an important role.
ResponseEntity allows us to return not only the body (in this case, a list of products), but also additional HTTP metadata like status codes and headers. This makes the API more flexible.

For example, we can return:
- ResponseEntity.ok(body) for 200 OK
- ResponseEntity.status(HttpStatus.NOT_FOUND).build() if something is missing
- ResponseEntity.badRequest() for errors.

ResponseEntity gives us more flexibility and is a good fit when designing APIs that communicate over HTTP.

src/main/kotlin/ecommerce/Product.kt
mariyakulikova: From what I understand, the term `entity` usually refers to an object that is meant to be persisted in a databaseю. It has a unique identity and represents a row in a table. 
Using a `toEntity()` method has a few advantages:
-  It encapsulates the logic of creating an entity. If we ever need to adjust how entities are built, we only need to change it in one place.
- We are explicitly converting some kind of data into a database entity.
- It avoids repeating the object construction logic throughout the codebase.

 In the current stage of the project, the database is responsible for generating the id (via `AUTO_INCREMENT`):
```
CREATE TABLE products (
                          id BIGINT AUTO_INCREMENT PRIMARY KEY,
                          name VARCHAR(255) NOT NULL,
                          price DOUBLE NOT NULL,
                          image_url VARCHAR(512) NOT NULL
);
```
I originally used `toEntity()` back when the database wasn't connected yet and the application was responsible for generating unique IDs manually. That’s why the method was introduced in the first place.

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: I added the `/api` prefix to make separation between API endpoints and potential non-API routes.

 -  `/api/products` - for frontend or external clients to fetch product data
 -  `/products` - render a view in a traditional web app

src/test/kotlin/ecommerce/DBTest.kt
mariyakulikova: We use lateinit here because we use `@Autowired` instead of constructor injection.
The `@Autowired` annotation tells Spring to automatically find and give us `ProductRepository`. We cannot initialize the variable ourselves, because Spring does it later when the test starts.

src/test/kotlin/ecommerce/DBTest.kt
mariyakulikova: It tells Spring to inject the needed dependency automatically.

src/test/kotlin/ecommerce/CRUDTest.kt
mariyakulikova: Thank you!

src/test/kotlin/ecommerce/CRUDTest.kt
mariyakulikova: I decided to drop and recreate the table in `@BeforeEach` to make sure each test starts with a clean DB. This way, tests don’t affect each other.
I wouldn't remove `@BeforeEach` even if you had asked 😆

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: Spring does all the magic here 😊
I don’t create the productRepository object myself. Spring sees the `@Repository` annotation on the `ProductRepository` class and automatically creates an instance during application startup. Then, it injects this instance into my controller using constructor injection.

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: After checking the difference, I’d choose `Unit`. Since we’re writing Kotlin code, `Unit` is more idiomatic, it’s a real object, and it integrates better with Kotlin features like expressions and generics. Also, `Void` comes from Java and can be confusing or nullable, while `Unit` is simple and safe.

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: Here’s what I found:
- `@Controller` is used for web pages
- `@RestController` is used for APIs
- Technically, `@RestController` = `@Controller` + `@ResponseBody`

I’ve replaced the `@Controller` annotation on `ProductController` with `@RestController`,
and introduced a `@RequestMapping("/api/products")` annotation at the class level to reduce duplication in method-level mappings.

src/test/kotlin/ecommerce/SSRTest.kt
mariyakulikova: I wanted to test here that the HTML page only displays products that actually exist. So checking that "melon" (which doesn’t exist in the DB) is not shown was meant to verify that.
I renamed the test to: `HTML page should display only the existing products`.

mariyakulikova: Now I see this test might not be needed 😄 I'm thinking of removing it. What do you think?

src/test/kotlin/ecommerce/SSRTest.kt
mariyakulikova: I added @AutoConfigureMockMvc by mistake 😅 at the beginning, I was trying many things and learning how everything works.
As far as I understand, this annotation is used when we want to use `MockMvc` for testing controllers without starting a full web server. But in this test, I don’t use `MockMvc`, so I think I can safely remove it.

src/test/kotlin/ecommerce/CRUDTest.kt
mariyakulikova: If we change the `@BeforeEach` logic, I could easily forget to update the magic number. I changed the test to check the number of products dynamically using the size of the list.
Thanks for pointing that out!

src/test/kotlin/ecommerce/CRUDTest.kt
mariyakulikova: I updated the test to use `RestAssured`’s built-in `assertThat` method. Now it’s much cleaner.
I’ll also check out more of its features in the docs. There’s definitely a lot I don’t know yet 😅

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: Here's what I learned after checking the docs:
- `ResponseEntity.created()`
Returns 201 Created.
Used when creating a new product. It also sets the Location header with a URI to the newly created resource.
- `ResponseEntity.ok()`
Returns 200 OK.
Used for successful GET requests and updates.
- `ResponseEntity.notFound()`
Returns 404 Not Found.
Used when a requested product is not found. For example, trying to update or delete a non-existent product.
- `ResponseEntity.noContent()`
Returns 204 No Content.
Used for successful DELETE operations when there’s no content to return.

build.gradle.kts
mariyakulikova: REST-assured was designed to simplify the testing and validation of REST APIs. It makes testing HTTP endpoints easier and more readable.
It helps us:
- send real HTTP requests in tests,
- check response status codes, headers, and bodies,
- log requests and responses for debugging.

src/main/kotlin/ecommerce/ProductController.kt
mariyakulikova: Here what I've learned:
- `@RequestBody` is used to map the body of the HTTP request to a Kotlin object. By default, it’s required, but we can make it optional with `@RequestBody(required = false)`.
- `@PathVariable` is used to extract values from the URI path `/products/{id}`. It also has an optional required parameter we can configure, like `@PathVariable(required = false)`.
- `Model` is used in MVC to pass data to the view template. It’s not related to HTTP request data, but used to render dynamic HTML pages.

Let me know if I misunderstood anything.

src/test/kotlin/ecommerce/SSRTest.kt
mariyakulikova: I was using `RestAssured` with `DEFINED_PORT`, which sends real HTTP requests. But I also had `@AutoConfigureMockMvc`, which is meant for `MockMvc` testing without a real server.
They don’t work well together, so I removed the annotation.

At first, it looked very hard and complicated, but step by step I broke it down.
Thank you for encouraging me!

src/test/kotlin/ecommerce/DBTest.kt
mariyakulikova: Here I wanted to test `Jdbc`-related code, so I added `@JdbcTest`. This annotation tells Spring to auto-configure `JdbcTemplate`.
For example, if I remove this annotation from the test class, the tests won’t run because of an injection error.
image

mariyakulikova:
Thank you for your detailed questions!
Preparing the answers really helped me understand the topics better and learn more deeply.
I appreciate the time you took to guide me. It was very helpful for my learning process 😊

src/main/kotlin/ecommerce/Product.kt
pkeugine: Nice explanation!
If I understood it correctly, it seems like `toEntity` and `update` method is not currently used.
It would be great if we could remove them from the project, and make one when we need it with logics that fit our needs in that situation.

src/main/kotlin/ecommerce/ProductController.kt
pkeugine: Perfect! It's the answer that I wanted 💯

src/main/kotlin/ecommerce/ProductController.kt
pkeugine: It's great 👍

src/main/kotlin/ecommerce/ProductController.kt
pkeugine: Yes 👍 
Spring framework registers them as `beans` and takes care of their lifecycle, such as : creation, wiring, and destruction. This allows us to utilize features that the framework provides when managing the bean, dependency injection being one of them. It would be nice to look up what other features Spring provides if you have the time :)

src/test/kotlin/ecommerce/CRUDTest.kt
pkeugine: I like the confidence 😄 
Yes, it's important to clean up the database if we're reusing it in multiple tests.
You've done a great job 👍

src/test/kotlin/ecommerce/DBTest.kt
pkeugine: Nice. But the tests pass as well if we use `@SpringBootTest`.
I was curious about what the differences were between `@SpringBootTest` and `@JdbcTest` 😄

src/test/kotlin/ecommerce/DBTest.kt
pkeugine: So we used constructor based dependency injection in application logic, and `@Autowired` for tests! 👍 
Are there any other ways for us to inject dependencies? 🤔

src/test/kotlin/ecommerce/SSRTest.kt
pkeugine: I agree. 👍

src/test/kotlin/ecommerce/CRUDTest.kt
```diff
@@ -0,0 +1,152 @@
...
+import io.restassured.http.ContentType
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.jdbc.core.JdbcTemplate
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
```
pkeugine, line 13: I'm surprised you noticed that `@DirtiesContext` annotation is not necessary here, since you're cleaning up the database before most of the tests 💯 

I didn't want to review on this annotation yet, since it requires some understanding about `ApplicationContext` in Spring. However, you managed to do it yourself 👍 

It would be great if we could remove the one in `SSRTest` as well.
Also, which approach do you think looks better : `@DirtiesContext` or `@BeforeEach` database cleanup?

src/test/kotlin/ecommerce/SSRTest.kt
mariyakulikova: 🙆‍♀️

src/main/kotlin/ecommerce/Product.kt
mariyakulikova: 🙆‍♀️

pkeugine (approved):
Hello Mariia, and thank you for the hard work answering all the questions 👍 

You've done a great job, and all the answers seem great.
I also noticed you've removed some unnecessary annotations as you were refactoring the code 💯 
I left some responses to your answers and follow-up questions, but there aren't that many. You can leave the answers here as comments, or put it in the next PR message!

I think we can safely go to the next step, so I'll merge this PR.
If you have any questions or opinions you'd like to share, feel free to DM me anytime.
Let's keep this momentum going, and as always,
Happy Hacking! 🔥 🔥 🔥

src/test/kotlin/ecommerce/DBTest.kt
mariyakulikova: Yes, there are three way to inject dependencies:
- Constructor injection,
- Field injection,
- Setter injection.

src/test/kotlin/ecommerce/DBTest.kt
mariyakulikova: The difference between `@SpringBootTest` and `@JdbcTest` is, that `@SpringBootTest` loads the entire application context, while `@JdbcTest` loads only the components needed for testing `Jdbc` database interactions. 🤓

src/test/kotlin/ecommerce/CRUDTest.kt
mariyakulikova: I noticed that `@DirtiesContext` was causing slower test runs, and once I understood what it does with the `ApplicationContext`, the cleanup approach felt more lightweight.
Between the two, I think I prefer `@BeforeEach` database cleanup. It’s explicit, faster, and keeps the context clean.

PR #4, opened 2025-07-17
[Spring-Ecommerce-Product Step 1] Submit Petra's mission

pebencze:
Hello!

Thank you for reviewing my work. This is my first project using the Spring framework. I have created a simple HTTP API that allows users to retrieve, add, update, and delete products. I also implemented a basic server-side rendering via HTML using Thymeleaf and JavaScript. As a third step, I have added a H2 database instead of using Kotlin collections.

After implementing these features, I still need to connect some dots and read more about Spring and SQL, so unfortunately I cannot ask many specific questions yet, but I am happy about any feedback. I am curious how you would test: using Postman, developer tools or using test frameworks and if yes, which ones do you recommend?

Best,

Petra

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Since no separate database is used at this point, data is stored in memory using
 - [x] resource representation class to store data
   - [x] is wrapped in a collection
 - [x] HTTP requests and responses in JSON format
-- [ ] retrieve product - GET
+- [x] retrieve product - GET
 - [x] add new product - POST
 - [ ] update product parameters - PUT
 - [ ] delete product from products - DELETE
--- a/README.md
+++ b/README.md
@@ -1,1 +1,15 @@
 # spring-ecommerce-product
+
+## Step 1-1
+Implements a simple HTTP API that allows users to **retrieve**, **add**, **update**, and **delete** products.
+HTTP requests and responses must be in **JSON** format.
+Since no separate database is used at this point, data is stored in memory using an appropriate Kotlin Collection Framework.
+
+### Feature list
+- [x] resource representation class to store data
+  - [ ] is wrapped in a collection
+- [ ] HTTP requests and responses in JSON format
+- [ ] retrieve - GET
+- [ ] add - POST
+- [ ] update - PUT
+- [ ] delete - DELETE
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Since no separate database is used at this point, data is stored in memory using
 
 ### Feature list
 - [x] resource representation class to store data
-  - [ ] is wrapped in a collection
-- [ ] HTTP requests and responses in JSON format
-- [ ] retrieve - GET
-- [ ] add - POST
-- [ ] update - PUT
-- [ ] delete - DELETE
+  - [x] is wrapped in a collection
+- [x] HTTP requests and responses in JSON format
+- [ ] retrieve product - GET
+- [x] add new product - POST
+- [ ] update product parameters - PUT
+- [ ] delete product from products - DELETE
--- a/README.md
+++ b/README.md
@@ -28,4 +28,16 @@ For product images, do not upload files; instead, use direct image URLs. <br/>
   - [x] based on HTML form submission and page navigation
   - [x] use direct image URLs for products
 
+## Step 1-3
+Refactor the application using an H2 in-memory database instead of Kotlin collections.
+The required database tables are initialized automatically when the application starts.
+
+### Feature List
+- [x] connect to a database
+  - [x] Add the required Gradle dependencies
+  - [x] Define the database schema
+  - [x] Configure the database settings
+  - [x] use Spring's JdbcTemplate
+
+
 
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ However, if you're interested in asynchronous behavior using JavaScript, feel fr
 For product images, do not upload files; instead, use direct image URLs. <br/>
 
 ### Feature list
-- [ ] admin interface
-  - [ ] uses Thymeleaf
-  - [ ] based on HTML form submission and page navigation
-  - [ ] use direct image URLs for products
+- [x] admin interface
+  - [x] uses Thymeleaf
+  - [x] based on HTML form submission and page navigation
+  - [x] use direct image URLs for products
 
 
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # spring-ecommerce-product
 
 ## Step 1-1
-Implements a simple HTTP API that allows users to **retrieve**, **add**, **update**, and **delete** products.
-HTTP requests and responses must be in **JSON** format.
-Since no separate database is used at this point, data is stored in memory using an appropriate Kotlin Collection Framework.
+Implements a simple HTTP API that allows users to **retrieve**, **add**, **update**, and **delete** products.<br/>
+HTTP requests and responses must be in **JSON** format. <br/>
+Since no separate database is used at this point, data is stored in memory using an appropriate Kotlin Collection Framework. <br/>
 
 ### Feature list
 - [x] resource representation class to store data
@@ -13,3 +13,19 @@ Since no separate database is used at this point, data is stored in memory using
 - [x] add new product - POST
 - [x] update product parameters - PUT
 - [x] delete product from products - DELETE
+
+## Step 1-2
+Implements an admin interface that allows users to view, add, update, and delete products. <br/>
+
+Use Thymeleaf to implement server-side rendering (SSR). <br/>
+The default behavior should be based on traditional HTML form submission and page navigation. <br/>
+However, if you're interested in asynchronous behavior using JavaScript, feel free to use the previously built product API to apply AJAX or similar techniques.<br/>
+For product images, do not upload files; instead, use direct image URLs. <br/>
+
+### Feature list
+- [ ] admin interface
+  - [ ] uses Thymeleaf
+  - [ ] based on HTML form submission and page navigation
+  - [ ] use direct image URLs for products
+
+
--- a/README.md
+++ b/README.md
@@ -47,6 +47,33 @@ The required database tables are initialized automatically when the application 
     - [x] Configure the database settings
     - [x] use Spring's JdbcTemplate
 
+## Learnings
+### JDBC & DAO
+- my `ProductRepository` class is a DAO (Data Access Object): a design pattern for data access abstraction, meaning that
+I don't expose how the products are stored and retrieved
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -1 +1,58 @@
...
+The default behavior should be based on traditional HTML form submission and page navigation. <br/>
+However, if you're interested in asynchronous behavior using JavaScript, feel free to use the previously built product
+API to apply AJAX or similar techniques.<br/>
+For product images, do not upload files; instead, use direct image URLs. <br/>
+
+### Feature list
+
+- [x] admin interface
+    - [x] uses Thymeleaf
+    - [x] based on HTML form submission and page navigation
```
Hyeon9mak, line 34: 👍 nice habbit

src/main/kotlin/ecommerce/Product.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce
+
+class Product(
+    var id: Long? = null,
+    var name: String = "",
+    var price: Double = 0.0,
+    var imageUrl: String = "",
+)
```
Hyeon9mak: In the context of entities, equality is determined based on the ID.

For example, if you have

- id = 1, name = "red apple"

On the other hand

```kotlin

        other as Product

        return id == other.id
    }

        return id?.hashCode() ?: 0
    }
```

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.ResponseBody
+import java.net.URI
+
+@Controller
+class ProductController(private val productRepository: ProductRepository) {
```
Hyeon9mak, line 16: and u can also append this annotation 😉 
what is it's role?

```kotlin
@RequestMapping("/products")
```

and what's different between `@RestController` & `@Controller` ? 😉

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+    @PutMapping("/products/{id}")
+    @ResponseBody
+    fun update(
+        @RequestBody newProduct: Product,
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        productRepository.edit(newProduct, id)
+        return ResponseEntity.ok().build()
+    }
+
```
Hyeon9mak: nice choice HTTP methods! 💯

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+    @DeleteMapping("/products/{id}")
+    @ResponseBody
+    fun delete(
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        productRepository.delete(id)
+        return ResponseEntity.noContent().build()
+    }
```
Hyeon9mak, line 52: 204 status code! nice 👍

src/main/kotlin/ecommerce/ProductRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+        productId: Long,
+    ) {
+        val sql = "UPDATE products SET name = ?, price = ?, image_url = ? WHERE id = ?"
+        jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, productId)
+    }
+
+    fun delete(id: Long): Int {
+        return jdbcTemplate.update("delete from products where id = ?", id)
+    }
+}
```
Hyeon9mak: if update or delete failed, how did we know that?

src/main/resources/schema.sql
```diff
@@ -0,0 +1,8 @@
+create table PRODUCTS
+(
+    id        int          not null AUTO_INCREMENT,
+    name      varchar(100) not null,
+    price     DECIMAL(10, 2),
+    image_url varchar(500) not null,
+    PRIMARY KEY (id)
+);
```
Hyeon9mak, line 8: there are name length, price format, image_url length limits!

I think adding validation logic to the `Product`, 
might make the code more robust and easier to test.

U think so too, right?
spring project also need OOP 😄

Hyeon9mak (changes requested):
Hi Petra! Welcome to the Spring mission. 🥳 

Implemented very nicely 👍 

As u know, the best testing approach often depends on the situation. 😄 
In Ur case, since u r currently learning various concepts centered around Spring Boot,
it would be best to focus on practicing with tools like JUnit and `@SpringBootTest`.

src/main/kotlin/ecommerce/Product.kt
pebencze: Very insightful. Thank you! Will implement this.

src/main/kotlin/ecommerce/ProductController.kt
pebencze: OK, thank you.` @RequestMapping` can be used on a class-level for shared mappings. I am wondering if it will conflict with /admin/products, but let's give it a try.
`@RestController `is a Controller that contains `@ResponseBody`, meaning that every method will return JSON if it returns a ResponseEntity. However, we have a MVC `@Controller` and one method returns a view that is not a JSON, but an HTML page. This is why I chose Controller. 
I could also split into ProductApiController and ProductViewController... shall I?

src/main/kotlin/ecommerce/ProductController.kt
pebencze: Put creates a new product if the requested product does not exist, patch only updates existing products. Maybe patch is more accurate for our purposes? Since we have create anyways.

src/main/kotlin/ecommerce/ProductRepository.kt
pebencze: Mh, we don't. I will look into it.

pebencze: I have changed update and delet functions, but I am not sure if and how I should handle failure during creation. It is quite unlikely that something fails, but if yes, I would throw 500? Then I need to change the return type to ResponseEntity<Any?> form Void...

src/main/resources/schema.sql
pebencze: Fair! This is going to be step two, so I will implement it in another branch. What do you thin of AUTO_INCREMENT for id? Is it better to use something like this for insertion aka creation?
`fun insertWithKeyHolder(customer: Customer): Long {
        val sql = "insert into customers (first_name, last_name) values (?, ?)"

        val keyHolder: KeyHolder = GeneratedKeyHolder()
        jdbcTemplate.update({
            it.prepareStatement(sql, arrayOf("id")).apply {
                setString(1, customer.firstName)
                setString(2, customer.lastName)
            }
        }, keyHolder)

        return keyHolder.key!!.toLong()
    }`

pebencze:
Hello @Hyeon9mak !

Thank you for the constructive feedback! I would appreciate if you could take a look at my changes and answer this question: I could also split into ProductApiController and ProductViewController... shall I?

Best,
Petra

src/main/kotlin/ecommerce/Product.kt
```diff
@@ -0,0 +1,21 @@
...
+        if (this == other) return true
+        if (other == null || javaClass != other.javaClass) return false
+
+        other as Product
+
+        return id != null && id == other.id
+    }
+
+    override fun hashCode(): Int {
+        return id?.hashCode() ?: System.identityHashCode(this)
```
Hyeon9mak, line 19: now if `Product` name(or price) changed,
we will be able to trace that it’s the same product, just with a different name! 👍

src/main/kotlin/ecommerce/ProductController.kt
Hyeon9mak: 💯 U already know it well.

> shall I?

Yeah! for practice and grow 😄

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,60 @@
...
+    @ResponseBody
+    fun update(
+        @RequestBody newProduct: Product,
+        @PathVariable id: Long,
+    ): ResponseEntity<Unit> {
+        if (!productRepository.update(newProduct, id))
+            return ResponseEntity.notFound().build()
+        return ResponseEntity.ok().build()
+    }
+
```
Hyeon9mak, line 43: > Put creates a new product if the requested product does not exist, patch only updates existing products. Maybe patch is more accurate for our purposes? Since we have create anyways.

Hmm, is that really the difference? 🤔 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/PATCH

src/main/resources/schema.sql
Hyeon9mak: > What do you thin of AUTO_INCREMENT for id? Is it better to use something like this for insertion aka creation?

Nice idea!
I think it's up to Petra and the team to decide how to proceed! 😄 

However, since the main focus of this mission is Spring and OOP,
I don't think there's a strong need to change it at this point. 👍

src/main/kotlin/ecommerce/ProductRepository.kt
```diff
@@ -0,0 +1,62 @@
...
+    ): Boolean {
+        val sql = "UPDATE products SET name = ?, price = ?, image_url = ? WHERE id = ?"
+        val rowsAffected = jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, productId)
+        return rowsAffected > 0
+    }
+
+    fun delete(id: Long): Boolean {
+        val rowsAffected = jdbcTemplate.update("delete from products where id = ?", id)
+        return rowsAffected > 0
+    }
```
Hyeon9mak, line 61: 👍 

> It is quite unlikely that something fails, but if yes, I would throw 500? Then I need to change the return type to ResponseEntity<Any?> form Void...

You're right. Since it's a technical issue on the server side and not the user's fault, returning a 500 error is the correct approach.
How about letting the controller only define the expected normal behavior, and handling the error using a separate Exception Handler? 
https://spring.io/blog/2013/11/01/exception-handling-in-spring-mvc 😉

Hyeon9mak (approved):
Hi Petra!

Great job incorporating the feedback!
I’ve left some answers to your questions, but it’s totally fine to go over them while working on the next step. 😄 

See u next step!
If you have any further questions, feel free to ask anytime!

PR #5, opened 2025-07-17
[spring-ecommerce-product - Step 1-1 to 1-3] Submit AnnLvu's mission

AnnLvu:
Hi! 
Hope you are doing great! 👋
It was my first encounter with the Spring Framework and databases. I hope we did well. I learned a lot of new things. I don't fully understand some aspects yet, but I will keep practicing and reading about the differences between @RestController and @Controller. I also need to dive deeper into SQL commands.

`Product` -  is our model
`ProductController` -  handles the CRUD operations
`ProductRepository` -  handles the SQL statements

I probably can't ask many questions yet. But I would like to hear your thoughts on our project. What would you suggest to improve or modernize?

I am curious how you would test: using Postman, developer tools or using test frameworks and if yes, which ones do you recommend?

Thanks so much for taking the time to look it over! 🙏

All the best,
Ann

### DIFF
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -49,7 +49,8 @@ class ProductController {
     @GetMapping("/admin/products")
     fun table(model: Model): String {
         model.addAttribute("products", products.values)
-        return "/products"
+        model.addAttribute("product", Product())
+        return "products"
     }
 }
 
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -1,0 +1,22 @@
+package ecommerce
+
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+import java.util.concurrent.atomic.AtomicLong
+
+@RestController
+class ProductController {
+    private val products: MutableMap<Long, Product> = HashMap()
+    private val index = AtomicLong(1)
+
+    @PostMapping("/products")
+    fun create(@RequestBody product: Product): ResponseEntity<Void> {
+        val newProduct = Product.toEntity(product, index.incrementAndGet())
+        val productId = newProduct.id ?: throw RuntimeException("Product id is null")
+        products.putIfAbsent(productId, newProduct)
+        return ResponseEntity.created(URI.create("/products/${newProduct.id}")).build()
+    }
+}
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -2,7 +2,9 @@ package ecommerce
 
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
 import java.net.URI
@@ -24,5 +26,16 @@ class ProductController {
     @GetMapping("/products")
     fun read(): ResponseEntity<MutableMap<Long, Product>> {
         return ResponseEntity.ok().body(products)
+    }
+
+    @PutMapping("/products/{id}")
+    fun update(@RequestBody newProduct: Product, @PathVariable id: Long): ResponseEntity<Void> {
+        if (!products.containsKey(id)) {
+            create(newProduct)
+            return ResponseEntity.ok().build()
+        }
+        val product = products[id] ?: throw RuntimeException("Product id is null")
+        product.update(newProduct)
+        return ResponseEntity.ok().build()
     }
 }
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -1,22 +1,19 @@
 package ecommerce
 
 import org.springframework.http.ResponseEntity
-import org.springframework.web.bind.annotation.DeleteMapping
-import org.springframework.web.bind.annotation.GetMapping
-import org.springframework.web.bind.annotation.PathVariable
-import org.springframework.web.bind.annotation.PostMapping
-import org.springframework.web.bind.annotation.PutMapping
-import org.springframework.web.bind.annotation.RequestBody
-import org.springframework.web.bind.annotation.RestController
+import org.springframework.stereotype.Controller
+import org.springframework.ui.Model
+import org.springframework.web.bind.annotation.*
 import java.net.URI
 import java.util.concurrent.atomic.AtomicLong
 
-@RestController
+@Controller
 class ProductController {
     private val products: MutableMap<Long, Product> = HashMap()
     private val index = AtomicLong(0)
 
     @PostMapping("/products")
+    @ResponseBody
     fun create(@RequestBody product: Product): ResponseEntity<Void> {
         val newProduct = Product.toEntity(product, index.incrementAndGet())
         val productId = newProduct.id ?: throw RuntimeException("Product id is null")
@@ -25,11 +22,13 @@ class ProductController {
     }
 
     @GetMapping("/products")
+    @ResponseBody
     fun read(): ResponseEntity<MutableMap<Long, Product>> {
         return ResponseEntity.ok().body(products)
     }
 
     @PutMapping("/products/{id}")
+    @ResponseBody
     fun update(@RequestBody newProduct: Product, @PathVariable id: Long): ResponseEntity<Void> {
         if (!products.containsKey(id)) {
             create(newProduct)
@@ -41,8 +40,18 @@ class ProductController {
     }
 
     @DeleteMapping("/products/{id}")
+    @ResponseBody
     fun delete(@PathVariable id: Long): ResponseEntity<Void> {
         products.remove(id) ?: throw RuntimeException("Product id is null")
         return ResponseEntity.noContent().build()
     }
+
+    @GetMapping("/admin/products")
+    fun table(model: Model): String {
+        model.addAttribute("products", products.values)
+        return "/products"
+    }
 }
+
+// TODO read about difference btw RestController and Controller
+// TODO why do we combine Controller and ControllerView
--- a/src/main/kotlin/ecommerce/ProductController.kt
+++ b/src/main/kotlin/ecommerce/ProductController.kt
@@ -3,23 +3,22 @@ package ecommerce
 import org.springframework.http.ResponseEntity
 import org.springframework.stereotype.Controller
 import org.springframework.ui.Model
-import org.springframework.web.bind.annotation.*
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.ResponseBody
 import java.net.URI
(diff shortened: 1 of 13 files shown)


src/test/kotlin/ecommerce/ProductControllerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.test.annotation.DirtiesContext
+
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
Wordbe: What was the reason for applying this annotation?
It seems to work fine even without it. 🤔

src/test/kotlin/ecommerce/ProductControllerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+            RestAssured.given().log().all().body(Product(name = "cola", price = 4.5, imageUrl = "https://cola.jpg"))
+                .contentType(ContentType.JSON).`when`().post("/products").then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.CREATED.value())
+    }
+
+    @Test
+    fun read() {
+        val products = productRepository.findAllProducts()
+        assertThat(products.size).isEqualTo(4)
```
Wordbe, line 60: It would be great if you could validate all the fields more thoroughly and specifically!

src/test/kotlin/ecommerce/ProductControllerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+class ProductControllerTest {
+    @Autowired
+    private lateinit var productRepository: ProductRepository
+
+    @Autowired
+    private lateinit var jdbcTemplate: JdbcTemplate
+
+    @BeforeEach
+    fun setUp() {
+        productRepository = ProductRepository(jdbcTemplate)
```
Wordbe: This code doesn't seem necessary —
shall we take a moment to figure out why that might be? 🙂

src/test/kotlin/ecommerce/ProductControllerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+                "tea 4 http//tea",
+            ).map { name -> name.split(" ").toTypedArray() }.toList()
+        jdbcTemplate.batchUpdate("INSERT INTO products(name, price, image_url) VALUES (?,?,?)", splitUpAttributes)
+    }
+
+    @Test
+    fun create() {
+        val response =
+            RestAssured.given().log().all().body(Product(name = "cola", price = 4.5, imageUrl = "https://cola.jpg"))
+                .contentType(ContentType.JSON).`when`().post("/products").then().log().all().extract()
```
Wordbe: It might be even better if you separate the lines a bit more to improve readability! 🙂

src/main/resources/schema.sql
```diff
@@ -0,0 +1,8 @@
+create table PRODUCTS
+(
+    id        int          not null AUTO_INCREMENT,
```
Wordbe: It would be a good idea to align the id type between the application and the database for consistency.
Shall we take a look at the potential issues that could arise if they don’t match? 🧐

src/main/kotlin/ecommerce/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.ResponseBody
+import java.net.URI
+
+@Controller
+class ProductController(private val productRepository: ProductRepository) {
```
Wordbe: Try separating the API Controller and the View Controller.
(Hint: use @RestController for APIs and @Controller for views)

src/test/kotlin/ecommerce/ProductControllerTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        jdbcTemplate.execute(
+            "CREATE TABLE products(" + "id SERIAL, name VARCHAR(100), price DECIMAL(10,2), image_url VARCHAR(500))",
+        )
+
+        val splitUpAttributes: List<Array<String>> =
+            listOf(
+                "cola 2 http//cola",
+                "fanta 3 http//fanta",
+                "coffee 4 http//coffee",
+                "tea 4 http//tea",
```
Wordbe: Try to make the data (especially the urls) as realistic as possible — it helps make the code more meaningful and testable!

src/main/kotlin/ecommerce/ProductRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+        productId: Long,
+    ) {
+        val sql = "UPDATE products SET name = ?, price = ?, image_url = ? WHERE id = ?"
+        jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, productId)
+    }
+
+    fun delete(id: Long): Int {
+        return jdbcTemplate.update("delete from products where id = ?", id)
+    }
+}
```
Wordbe: The DML keywords are mixed with both uppercase and lowercase — it would be better to standardize them to one style.

src/main/resources/application.properties
```diff
@@ -1 +1,9 @@
 spring.application.name=spring-ecommerce
+spring.h2.console.enabled=true
+spring.h2.console.path=/h2-console
+spring.datasource.url=jdbc:h2:mem:database
+spring.jpa.defer-datasource-initialization=true
```
Wordbe: What’s the purpose of this setting? 🤔

Wordbe (changes requested):
Hello Ann!

Welcome to your first Spring project mission.

I liked that you made the features work and included tests for the Controller.
How about adding some tests for the `ProductRepository` as well?

I’ve also left additional review comments—please take a look and apply any changes you can.
Thank you! 😊

src/test/kotlin/ecommerce/ProductControllerTest.kt
AnnLvu: I fully agree, now I added a check for compliance with the product list!

src/test/kotlin/ecommerce/ProductControllerTest.kt
AnnLvu: I added real names, prices and URL. Thank you so much! Now it looks more intuitive

src/main/resources/schema.sql
AnnLvu: **Issue** : If I use just `INT` for the `ID` column in the `PRODUCTS` table limits the range to `[MIN_INT; MAX_INT]` . In the application, we use Long (64-bit, range `[Long.MIN_VALUE ; Long.MAX_VALUE`]), which could cause **overflow errors** if the ID exceeds INT's limit .

**Solution**: I replaced INT with BIGINT for the ID column to match the Long type in the application, ensuring compatibility and supporting a larger range of values without overflow risks.

Thank you for the note! It was my stupid mistake :)

src/main/kotlin/ecommerce/ProductRepository.kt
AnnLvu: Replaced all DML keywords with uppercase.  Now it looks better!

src/test/kotlin/ecommerce/ProductControllerTest.kt
AnnLvu: Hi @Wordbe! 
I don't fully understand what do you mean here:
- that's only about line 25
- or about fun setUp()?

AnnLvu: I understand it like this: the line` productRepository = ProductRepository(jdbcTemplate)` is redundant since @Autowired already injects the ProductRepository bean, configured by Spring with the correct JdbcTemplate. Is that correct?

src/test/kotlin/ecommerce/ProductControllerTest.kt
AnnLvu: @Wordbe 
I believe you’re suggesting that `@DirtiesContext(classMode = BEFORE_EACH_TEST_METHOD)` might not be needed since the `setUp `method’s `DROP TABLE IF EXISTS PRODUCTS` ensures database isolation for each test. Is that what you meant? I’ve removed `@DirtiesContext` to speed up test execution, and all tests pass as `ProductRepository` doesn’t cache data or hold state. Since if  I planning to add more tests in the future, should I keep `@DirtiesContext` for robustness, or is the database reset sufficient? I’m also considering using `TRUNCATE TABLE` instead of` DROP TABLE` for efficiency - does that align with best practices?

src/main/kotlin/ecommerce/ProductController.kt
AnnLvu: Thank you for your suggestion to separate the _API_ and _view logi_c in `ProductController`. 
I've split it into two controllers: `ProductApiController` (annotated with `@RestController`) for handling REST API endpoints and `ProductViewController` (annotated with `@Controller`) for rendering view.

src/main/resources/application.properties
AnnLvu: Thank you for your question about the settings in `application.properties`! Here’s an explanation of each:

- `spring.h2.console.enabled=true`: enables the H2 web console, which allows viewing and testing the database via a browser
- `spring.h2.console.path=/h2-console`: specifies the URL path for accessing the H2 console
- `spring.datasource.url=jdbc:h2:mem:database`: configures the use of an in-memory H2 database named database
- `spring.jpa.defer-datasource-initialization=true`: delays the execution of schema.sql and data.sql until Hibernate completes schema initialization

Since our project uses JdbcTemplate (in ProductRepository) without Spring Data JPA, and the database schema is fully managed by schema.sql, this setting is unnecessary.
But I accessed the h2-consolee to explore its interface and experiment with SQL queries, which helped me better understand the database structure and data.

src/test/kotlin/ecommerce/ProductControllerTest.kt
```diff
@@ -0,0 +1,133 @@
...
+                    "adobestock_219364830-1440x700-crop-42-46.jpg",
+            )
+        }
+        with(products[3]) {
+            assertThat(id).isEqualTo(4)
+            assertThat(name).isEqualTo("Tea")
+            assertThat(price).isEqualTo(1.59)
+            assertThat(imageUrl).isEqualTo("https://cupitol.com/wp-content/uploads/2019/08/tea-drinking-1.jpg")
+        }
+    }
```
Wordbe, line 95: It looks good to override the `equals` and `hashCode` methods to compare objects!
(Reference: `data class`) 😊

src/test/kotlin/ecommerce/ProductControllerTest.kt
Wordbe: That's correct, yes exactly! 😊

src/main/resources/schema.sql
Wordbe: Great Job!

src/test/kotlin/ecommerce/ProductControllerTest.kt
Wordbe: Yes, in this test, `@BeforeEach` and `@DirtiesContext` seem to have overlapping roles. I’m thinking of using `@BeforeEach` for setting up data, and then considering `@DirtiesContext` only when needed!

Also, `TRUNCATE` looks like a good option! 😊
How about we look into its advantages and give it a try?

Wordbe (approved):
Hello Ann!
I can see you've carefully reviewed my comments and reflected on them thoughtfully. 🙂👍
I appreciated the relevant follow-up questions and your suggestion of new ideas like using `TRUNCATE`.

Just one thing — please make sure to add a few more tests, such as for `ProductRepositoryTest`, to strengthen the test coverage.

I'll approve and merge this PR as it stands now. Please reflect the above points and the ones in the comments in the next step. 

Great work, and thank you!

PR #6, opened 2025-07-17
[Spring-Ecommerce-Product Step 1] Submit saragrosser' mission

saragrosser:
This mission was a meaningful learning curve for me as I deepened my understanding of Spring Boot and backend fundamentals. While implementing the different steps, I gradually moved from simple in-memory storage to a more realistic setup with database integration, all while reinforcing the concepts of REST APIs, server-side rendering, and separation of concerns.

### Key Takeaways & Reflections
- Understanding CRUD at a deeper level:
Creating endpoints for basic Create, Read, Update, and Delete operations helped me internalize the REST principles and how to apply them using Spring annotations. It was especially interesting to see how small changes in the structure of my controller could affect the flow and maintainability of the application.

- From JSON APIs to server-side rendering:
Step 1-2 was a major shift in perspective. Moving from JSON responses to HTML templates with Thymeleaf gave me insight into how traditional server-rendered applications work. I learned how form submissions and URL mappings are handled on the backend and how data flows from controller to template.

- Introducing the database (H2):
Step 1-3 brought everything closer to a real-world backend scenario. Replacing Kotlin collections with an H2 database helped me understand the basics of schema management. Setting up JdbcTemplate, wiring the repository layer, and writing SQL scripts allowed me to see how logic, data, and configuration come together in Spring-based apps.

Overall, this mission helped me appreciate the structure and power of the Spring framework. It also highlighted how important it is to keep each layer clean and focused

Thank you for taking the time to review this PR — any feedback is welcome and appreciated!


src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+class ProductControllerTest {
```
pkeugine, line 15: The tests are failing due to different kinds of reasons:
Screenshot 2025-07-19 at 3 16 40 AM

Please make sure to check that
* all tests pass
* project builds successfully

before creating a pull request.

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,69 @@
...
+
+@Controller
+@RequestMapping("/products")
+class ProductController {
+    @Autowired
+    private lateinit var productRepository: ProductRepository
+
+    private val products: MutableSet<Product> = HashSet()
+
+    private var index: AtomicLong = AtomicLong(1)
```
pkeugine: Since we're using a repository, let's remove these unused codes.
It would be nice to check tests as well and see if there are any.

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+import java.util.concurrent.atomic.AtomicLong
+
+@Controller
+@RequestMapping("/products")
```
pkeugine: I see you've used `@Controller` and `@RequestMapping` annotations!
What do these annotations do?

There's also `@RestController` annotation, so let's compare these annotations and see which one fits our needs. It would be great if you could tell me what you've researched 👍

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,69 @@
...
+class ProductController {
+    @Autowired
+    private lateinit var productRepository: ProductRepository
```
pkeugine: In order for us to use `productRepository`, you've added `@Autowired` + `lateinit` combo.
What do they do? Also, could there be any other way of achieving the same goal?
By 'goal', I mean using this `productRepository` object :)

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,69 @@
...
+
+    private val products: MutableSet<Product> = HashSet()
+
+    private var index: AtomicLong = AtomicLong(1)
+
+    @PostMapping(consumes = ["application/json"])
```
pkeugine: It's interesting that this method (api) consumes `application/json`.
What is this `application/json`? Are there any other type of data that an api can consume?
What about producing? I mean, why do we explicitly code `consumes` but not what it produces?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,69 @@
...
+        return "table"
+    }
+
+    @GetMapping("/{id}")
+    fun getProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> {
+        val product = productRepository.findById(id) ?: throw NoSuchElementException()
+        return ResponseEntity.ok(product)
+    }
```
pkeugine: One method returns a string, while some return a `ResponseEntity`.
What's the difference?
Also, I would like you to tell me what this `ResponseEntity` is and why you used it :)

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,51 @@
+package ecommerce.repository
+
+import ecommerce.model.Product
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.stereotype.Repository
+import java.sql.ResultSet
+
+@Repository
+class ProductRepository(private val jdbcTemplate: JdbcTemplate) {
```
pkeugine, line 10: We've seen `@Controller` and now here's `@Repository`.
What do you think are their similarities and differences?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+        return jdbcTemplate.queryForObject("select id, name, price, image_url from products where id = ?", productRowMapper, id)
+    }
+
+    fun getAll(): List<Product> {
+        return jdbcTemplate.query("select id, name, price, image_url from products", productRowMapper)
+    }
+
+    fun createProduct(product: Product) {
+        jdbcTemplate.update(
+            "insert into products (name, price, image_url) values (?,?,?)",
```
pkeugine: It would be great if we keep the convention of keeping SQL keywords all uppercase just like you did to `UPDATE` :
```sql
INSERT INTO products (name, price, image_url) VALUES (?,?,?)
```

It makes reading queries a bit easier, and keeping conventions makes the code look more... professional 😎

src/main/resources/application.properties
```diff
@@ -1 +1,11 @@
...
+
+spring.h2.console.enabled=true
+spring.h2.console.path=/h2-console
+spring.datasource.url=jdbc:h2:mem:database
+
+
+spring.sql.init.mode=always
+
+#spring.sql.init.schema-locations=classpath:sqlSchema/schema.sql
+#spring.sql.init.data-locations=classpath:sqlSchema/data.sql
```
pkeugine: It would be great if you could tell me what these properties are, and why the last two were commented out!

src/main/resources/schema.sql
```diff
@@ -0,0 +1,8 @@
+create table PRODUCTS
+(
+    ID       int              not null AUTO_INCREMENT,
+    NAME     varchar(100)     not null,
+    PRICE    double not null,
+    IMAGE_URL varchar(500),
+    PRIMARY KEY (ID)
+);
```
pkeugine:
```suggestion
CREATE TABLE products
(
    id        INT NOT NULL auto_increment,
    name      VARCHAR(100) NOT NULL,
    price     DOUBLE NOT NULL,
    image_url VARCHAR(500),
    PRIMARY KEY (id)
);
```

Again, conventions and formatting for professionalism 😎

src/main/resources/templates/table.html
```diff
@@ -0,0 +1,183 @@
+<!DOCTYPE html>
```
pkeugine, line 1: [I see you've applied TailwindCSS.](https://github.com/herotechcourse/spring-ecommerce-product/pull/6/commits/6f0c99d1ccd0bd7eee9c70e91906f44abfbde77c) 😮 
I'm guessing you or the pair had some experience with it?

build.gradle.kts
```diff
@@ -23,12 +23,14 @@ dependencies {
...
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
     implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("nz.net.ultraq.thymeleaf:thymeleaf-layout-dialect:3.2.0")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     runtimeOnly("com.h2database:h2")
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("io.rest-assured:rest-assured:5.3.1")
```
pkeugine, line 33: In our previous missions, we didn't use these two additional libraries.
What do you think these libraries allow us to do?
I ask this because I think knowing our tools and why we need it is as important as knowing how to use it :)

pkeugine (changes requested):
Hello Sari, and welcome to the world of Spring (in the middle of summer 🌞 ) !

It was fun reading your journey on this mission, and it reminded me of my experience of learning this framework :)

I left some comments, mostly questions, since I thought these questions will really help you deepen your understanding in Spring & Spring Boot. There is a very important comment that I would like you to check, and I recommend answering all the questions! We'll go over more once we're done with them.

If you have any questions, feel free to leave comments or DM me anytime.
Happy Hacking! 🏃 🏃 🏃

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
saragrosser: Hi pkeugine! 

I apologize for this mistake, I guess we were running at the end and didn't updated the tests, this refactor actually made me change a lot of the structure of the code so I am quite happy, it is starting to make more sense in terms of clean separation

 I focused on separating the logic between the API and the UI in line with SRP and best practices:

- I moved the REST logic to a new `ProductApiController` under the `/api/products` path, so that it handles all JSON responses cleanly.
- The original `ProductController` now only serves the Thymeleaf-based HTML views (`/products`).
- I updated the test class `ProductControllerTest` to use the `/api/products` endpoints and made it more robust by dynamically handling product IDs.
- This resolves the issues where the tests were failing due to ID collisions (with `data.sql`) and incorrect response formats (HTML vs JSON).

src/main/kotlin/ecommerce/controller/ProductController.kt
saragrosser: I have done this in the last refactoring process now ProductAdminController has a clean code without unused codes  :)

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
saragrosser: I forgot to mention that I also refactor the names so it makes more sense in terms of responsability handle: 

- ProductAdminController -> /products -> HTML 
- ProductApiController -> api/products -> JSON

src/main/kotlin/ecommerce/controller/ProductController.kt
saragrosser: Thank you for pointing this out! 
I learned that @Controller is used when returning HTML views, while @RestController is a combination of @Controller + @ResponseBody, and is meant for APIs that return JSON or other data directly.

When to use which?: 
@Controller -> to return HTML pages and when building a web app 
@ResponseBody -> to return JSON or raw data and when building a REST API 

With my new refactoring I also address this point by separating the logic

src/main/kotlin/ecommerce/controller/ProductController.kt
saragrosser: The @Autowired annotation tells Spring to inject the dependency automatically, and I used lateinit because the property is initialized after the object is created. 
However, after studying I’ve also learned that a more modern and preferred way in Kotlin is using constructor injection, which would let me avoid lateinit and write cleaner code

So I have refactored it to replace field injection with constructor injection in my ProductAdminController
- Injected ProductRepository via primary constructor
- Aligned with Kotlin + Spring best practices

src/main/kotlin/ecommerce/controller/ProductController.kt
saragrosser: In the earlier version of this controller, that was a way to clarify that the endpoint was strictly meant to receive JSON data in the request body (which matched the tests and API use case at that point).

However, since then I've refactored the project to cleanly separate the API and the UI logic:

/api/products -> is now handled by ProductApiController, which uses @RestController and leverages Spring’s default behavior for both consumes and produces as application/json. In those endpoints, I don’t manually specify either unless needed,Spring infers them based on annotations and content negotiation.

/products -> is now handled by ProductAdminController, which uses @Controller + Thymeleaf to return HTML. The @PostMapping(consumes = ["application/json"]) was a leftover from earlier iterations and is actually unnecessary now, since this controller isn’t the main entry point for JSON requests anymore.

To answer your point:

-consumes = "application/json" defines what content type the server accepts in the request body
-produces = "application/json" defines what format the server returns in the response

In the @RestController case, JSON is both default input and output, so we don’t usually need to specify either unless we want to support additional formats like XML, plain text, form data, etc right? 

Do you usually prefer making produces explicit for clarity, even when it's the default?

src/main/kotlin/ecommerce/controller/ProductController.kt
saragrosser: This was my first time doing this so we mixed a lot of logic, from what I’ve learned now, the difference comes down to how Spring handles the return type depending on the controller’s purpose:

-When returning a String in a @Controller, Spring treats it as the name of a view to render. For example, in the getProducts() method, "table" refers to the Thymeleaf template that displays the product list in HTML.

-In contrast, when returning a ResponseEntity, Spring sends a full HTTP response — including status code, headers, and body. This is much more suited to REST APIs where we need to control exactly what is returned to the client (e.g., 200 OK vs 404 Not Found, or custom headers).

Initially, I had both types mixed in one controller because I was handling both HTML rendering and JSON APIs together. After refactoring, I split the responsibilities:

- ProductAdminController (HTML UI) returns String view names

- ProductApiController (REST API) returns ResponseEntity<Product> or ResponseEntity<List<Product>>

src/main/kotlin/ecommerce/repository/ProductRepository.kt
saragrosser: Both  help Spring manage the class as a component and allow for dependency injection, but they serve different roles: @Controller handles web requests and routes them, while @Repository handles data access and persistence logic.
I now understand that these annotations also help with exception handling and readability by clearly showing each class's responsibility in the application.

src/main/resources/application.properties
saragrosser: These properties are used to automatically run SQL scripts during application startup. 
- spring.sql.init.mode=always ensures that schema/data files are processed, while the commented lines point to the expected locations of the SQL files. 

I commented them out because we were not using them and forgot to delete them, I have deleted them now because I am currently not using them

src/main/resources/templates/table.html
saragrosser: This was my first time hearing about TailwindCSS my pair implemented it for the style purposes :)

build.gradle.kts
saragrosser: Thank you for this comment that made me study more about libraries! 

The two dependencies  added serve specific roles: 
- thymeleaf-layout-dialect helps with layout templating for Thymeleaf, making it easier to reuse components like headers or footers in HTML views. 
- The jackson-module-kotlin dependency is important for serializing/deserializing Kotlin data classes correctly when dealing with JSON in REST APIs.

src/main/kotlin/ecommerce/repository/ProductRepository.kt
saragrosser: Thank you for this feedback I have refactored the code to follow SQL conventions, it is always nice to make the code more professional and to follow the conventions I will keep this in mind for the following steps!

src/main/resources/schema.sql
saragrosser: DONE thank you :)

src/main/kotlin/ecommerce/controller/ProductController.kt
pkeugine: This is a tip for writing PR comments with annotations:

GitHub actually used `@` for mentioning users like : @pkeugine 
And interestingly, there are users with names like `RestController`, and `responsebody`... 🙄 

So when writing comments about annotations, it's good to use the ` (backtick) and wrap around them to make it look like `@RestController`!

src/main/kotlin/ecommerce/controller/ProductController.kt
pkeugine: Perfect!
It would be great to know why constructor based injection is preferred as well. :)

src/main/kotlin/ecommerce/controller/ProductController.kt
pkeugine: Great answer.
Just like your conclusion, I don't usually explicitly specify the content types.
I mark them when I have to change it!
I thought this was a nice time for us to go over what `consumes` and `produces` are, and see what kind of content types there are. :)

src/main/resources/templates/table.html
pkeugine: It's always good to see some exploration in technology 👍

build.gradle.kts
pkeugine: Nice!
I actually left this comment for explanation for `thymeleaf-layout-dialect` and `rest-assured`,
but I think the answer is still great. 👍

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,74 @@
...
+                .given().log().all().body(product)
+                .contentType(ContentType.JSON)
+                .`when`().post("/api/products")
+                .then().extract().response()
+
+        assertThat(response.statusCode).isEqualTo(HttpStatus.OK.value())
+        val created = response.`as`(Product::class.java)
+        assertThat(created.name).isEqualTo("table")
+        assertThat(created.price).isEqualTo(45.0)
+        assertThat(created.imageUrl).isEqualTo("test.com")
```
pkeugine, line 45:
```suggestion
        RestAssured
            .given().log().all().body(product)
            .contentType(ContentType.JSON)
            .`when`().post("/api/products")
            .then()
            .statusCode(HttpStatus.OK.value())
            .body("name", equalTo("table"))
            .body("price", equalTo(45.0F))
            .body("imageUrl", equalTo("test.com"))
```
This is not a must, but we can also let `RestAssured` assert everything :)

pkeugine (approved):
Hello Sari, and I'm surprised by your detailed answers to my questions. 👍 
It's nice to see that you've applied the reviews as well.

I left some responses to the answers and a review about `RestAssured`.
We are now ready to move on to the next step, so I'll merge this PR! :)

If you have any questions or opinions, feel free to DM me anytime.
I'll see you soon in the next PR, and until then,
Happy Hacking! 👍 👍 👍

PR #7, opened 2025-07-17
[spring-ecommerce-product - Step1] Submit san-ghun's mission

san-ghun:
# Description

This PR is an implementation of a simple HTTP API that allows users to retrieve, add, update, and delete products using Spring Boot framework with Kotlin, also using H2 in-memory database for development purpose. It provide access to Product API with basic admin-like views to list products, add a product, update a product, and delete a product. It only include API for Product, using `ProductController`, `ProductService`, `ProductViewController`, and `Product`.

# Goals

- [x] Run and test the web application on local machine (localhost)
    - [x] Access application pages through a web browser and test the features.

- [x] Receive and respond to web requests using the Spring Boot framework
    - [x] Map controller methods to incoming HTTP requests based on URL and HTTP method.
    - [x] Use annotations to bind request parameters and request body to method arguments.
    - [x] Use a template engine to return dynamic web pages as responses.

- [x] Store and retrieve data using a database in a web application
    - [x] Configure DB connection settings in `application.properties`.
    - [x] Use `JdbcTemplate` to execute SELECT, INSERT, UPDATE, and DELETE queries.
    - [x] Understand what an in-memory DB is and how it can be used during development.

- [x] Understand and implement the HTTP protocol
    - [x] Understand the components of HTTP such as methods, request/response headers, and status codes.
    - [ ] Explain the purpose and differences between HTTP methods (GET, POST, PUT, DELETE).  *--still studying*
    - [ ] Understand and apply appropriate HTTP status codes (e.g., 200, 404, 500) in the correct situations.  *--still figuring out*

# Changes

- `ProductController` is a controller to provide requested data with response through API endpoints. It provide simple CRUD HTTP API to users.
- `ProductService` is a service layer to provide logic to support the `ProductController`, but also directly interact with the database.
- `ProductViewController` is a view controller to provide template view for web browser to show.
- Add configuration for H2 in-memory database as development purposed database, with sql file for schema and sample data.

# Testable Endpoints

- `POST /api/products` - create a product
- `GET /api/products` - read all products
- `GET /api/products/{id}` - read a product with id
- `PUT /api/products/{id}` - update a product with id
- `DELETE /api/products/{id}` - delete a product with id
- `GET /` - provide view of list of all products
- `GET /add/product` - provide view of form to add new product
- `GET /update/product/{id}` - provide view of form to update a product

# Additional Note

- Since the current `ProductService` is providing direct interaction with database, using the Data Access Object (DAO). I think it might better to change it into `ProductRepository`. Hope to know dear review's idea.

Thank you for your review!


README.md
```diff
@@ -1 +1,88 @@
...
+### Domain
+
+Product
+
+- [x] is a class
+- [x] has property `id: Long`
+    - [x] get `id` from controller with `AtomicLong(1)`
+- [x] has property `name: String`
+- [x] has property `price: Double`
+- [x] has property `imageUrl: String`
```
pkeugine, line 24: Nice job maintaining README 👍 
Looks clean and is very easy to read!

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
+@RequestMapping("/api")
```
pkeugine, line 15: I see you've used both `@RestController` and `@Controller` in your controllers!
What are the differences between these two annotations?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+    @GetMapping("/products")
+    fun getProducts(): ResponseEntity<List<Product>> {
+        val products = productService.findAll()
+        return ResponseEntity.ok(products)
+    }
+
+    @GetMapping("/products/{id}")
+    fun getProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> {
```
pkeugine, line 35: Nice work using `ResponseEntity` as a return type in controller.
In our previous missions, we would call `OutputView` since our objective was to print the info in the console. What is the role of this `ResponseEntity`? Are there any other options we can choose from?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+        return ResponseEntity.ok().build()
+    }
+
+    @DeleteMapping("/products/{id}")
+    fun deleteProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        productService.findById(id) ?: return ResponseEntity.notFound().build()
+        productService.delete(id)
+        return ResponseEntity.noContent().build()
```
pkeugine, line 59: `ResponseEntity.created`, `ResponseEntity.ok`, `ResponseEntity.notFound()`, `ResponseEntity.noContent`... So many different phrases related to http status!

I read the PR message and there was a checkbox about `Understand and apply appropriate HTTP status codes (e.g., 200, 404, 500) in the correct situations.`
 It'll be nice for you to study these status codes in detail and tell me about what you learned. :)

src/main/kotlin/ecommerce/api/Product.kt
```diff
@@ -0,0 +1,23 @@
...
+    var imageUrl: String,
+) {
+    fun update(newProduct: Product) {
+        this.name = newProduct.name
+        this.price = newProduct.price
+        this.imageUrl = newProduct.imageUrl
+    }
+
+    companion object {
+        fun toEntity(
```
pkeugine, line 10: What do you think this term`entity` means?
Also, when we want to set an id to a `product` we can simply:
```kotlin
        val product = Product(
            id = id,
            name = "product",
            price = 99.99,
            imageUrl = "https://example.com/sample.png",
        )
```
Why would we want to use this `toEntity` method instead, and should we enforce it? 🤔

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
...
+            Product(
+                rs.getLong("id"),
+                rs.getString("name"),
+                rs.getDouble("price"),
+                rs.getString("imageUrl"),
+            )
+        }
+
+    fun findAll(): List<Product> {
+        val products = db.query("""select * from product;""", productRowMapper)
```
pkeugine: I'm curious about why you used multi-line string for SQL!

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
...
+                    """select id, name, price, imageUrl from product where id = ?""",
+                    productRowMapper,
+                    id,
+                )
+        } catch (exception: Exception) {
+            println(exception.message)
+            return null
+        }
+        return product
+    }
```
pkeugine: I think it'll be simpler for us to return the result right away.
```suggestion
    fun findById(id: Long): Product? {
        try {
            return db.queryForObject(
                    """select id, name, price, imageUrl from product where id = ?""",
                    productRowMapper,
                    id,
                )
        } catch (exception: Exception) {
            println(exception.message)
            return null
        }
    }
```

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
...
+            product =
+                db.queryForObject(
+                    """select id, name, price, imageUrl from product where id = ?""",
+                    productRowMapper,
+                    id,
+                )
+        } catch (exception: Exception) {
+            println(exception.message)
+            return null
+        }
```
pkeugine: It's interesting to see `println` followed by `return null` when an exception gets thrown in database query logic. I want to know the thought process behind this!

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
...
+            product.imageUrl,
+        )
+    }
+
+    fun update(
+        id: Long,
+        product: Product,
+    ) {
+        db.update(
+            """UPDATE product set name = ?, price = ?, imageUrl = ? WHERE id = ?""",
```
pkeugine: Just like how you've wrote `INSERT INTO`, why don't we keep all the SQL keywords fully uppercased? :  
```sql
UPDATE product SET name = ?, price = ?, imageUrl = ? WHERE id = ?
```

Keeping this convention makes reading queries a bit easier, and makes the code look more... professional 😎

src/main/resources/application.properties
```diff
@@ -1 +1,9 @@
 spring.application.name=spring-ecommerce
+spring.h2.console.enabled=true
+spring.h2.console.path=/h2-console
+spring.datasource.url=jdbc:h2:mem:database
+#spring.jpa.defer-datasource-initialization=true
+#spring.jpa.hibernate.ddl-auto=none
+#spring.sql.init.mode=always
+spring.sql.init.schema-locations=classpath:schema.sql
+spring.sql.init.data-locations=classpath:data.sql
```
pkeugine, line 9: Can you explain to me what these properties do, and why some of them are commented out?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
+@RequestMapping("/api")
+class ProductController(private val productService: ProductService) {
+    @PostMapping("/products")
+    fun createProduct(
+        @RequestBody product: Product,
+    ): ResponseEntity<Void> {
```
pkeugine:
```suggestion
    ): ResponseEntity<Unit> {
```
This is not a must, but we can actually use `Unit` here as well.
What are the differences?
+) after looking up the differences, what would you use?

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
+package ecommerce.api
+
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.stereotype.Service
+import java.sql.ResultSet
+
+@Service
```
pkeugine: `@Service` annotation seems new. What does this do, and what made you choose this annotation?

pkeugine: I guess this question can be followed by your question:
**Q. Since the current ProductService is providing direct interaction with database, using the Data Access Object (DAO). I think it might better to change it into ProductRepository.**

But before we talk about rather `ProductService` is a great name or not,
it'll be great for us to go over what `service`, `DAO`, and `Repository` are.
What are these, and what made you think that `ProductRepository` could be a better fit?

src/main/resources/static/static.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <meta charset="UTF-8">
+    <title>Hello World Spring</title>
+</head>
+<body>
+Hello World!
+</body>
+</html>
```
pkeugine: Screenshot 2025-07-19 at 12 57 56 PM

Let's not forget to add an empty line at the end of all the files we try to merge.
Please check other files in the project as well.

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        jdbcTemplate.execute("DROP TABLE product IF EXISTS")
+        jdbcTemplate.execute(
+            """CREATE TABLE product (
+                         id          LONG    NOT NULL AUTO_INCREMENT,
+                         name        varchar(255)    NOT NULL,
+                         price       DOUBLE  NOT NULL,
+                         imageUrl    TEXT    NOT NULL,
+                         PRIMARY KEY (id)
+                    );""",
+        )
```
pkeugine: Why should we drop the table and create the same one **`BeforeEach` tests**?
I mean, we might as well reuse the one that we created before and save some time 🤔 

Oh, I'm not suggesting removal. I actually think it's great.
I'm asking this because I used to throw this question at myself when creating tests. 🙂

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.boot.test.autoconfigure.jdbc.JdbcTest
+import org.springframework.http.HttpStatus
+import org.springframework.jdbc.core.JdbcTemplate
+
+@JdbcTest
+class ProductControllerTest {
+    private lateinit var productService: ProductService
+
+    @Autowired
+    private lateinit var jdbcTemplate: JdbcTemplate
```
pkeugine: `lateinit` and `@Autowired` seems interesting 🤔 
Why should we use these?

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
```diff
@@ -0,0 +1,91 @@
...
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.JdbcTest
+import org.springframework.http.HttpStatus
+import org.springframework.jdbc.core.JdbcTemplate
+
+@JdbcTest
```
pkeugine, line 11: And yet, another annotation 😂 
I see you've chosen `@JdbcTest` as your way to test the project.
What does this annotation do, and why did you choose this?
Also, are there any other options we can choose from?

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
```diff
@@ -0,0 +1,91 @@
...
+        assertThat(response.statusCode.value()).isEqualTo(HttpStatus.NOT_FOUND.value())
+    }
+
+    @Test
+    fun update() {
+        val product = Product(name = "new product", price = 1.6, imageUrl = "https://www.product.com/image/2")
+        create()
+        val response = controller.updateProduct(1, product)
+        assertThat(response.statusCode.value()).isEqualTo(HttpStatus.OK.value())
+    }
```
pkeugine: I think there needs to be more than just `HttpStatus` assertion when it comes to testing api results, especially when it's related to `update`. However, the current update api doesn't allow us to check this with return value, since it's `ResponseEntity<Void>`. How could we check this?

pkeugine: Interesting topic:
If I attempt to update a product but the values are identical to the current ones,
is it appropriate for the server to respond with `200 OK`? It'll be fun to talk about his with fellow crews 🙂 

Also, I would like to know what you learned about:
**Q. Explain the purpose and differences between HTTP methods (GET, POST, PUT, DELETE).**
which you wrote in the PR message.

pkeugine (changes requested):
Hello San! 🙌 
Seems like you're on fire in this Spring Boot learning process 🔥 

In this first review, I focused mostly on asking you questions about `why` you used certain annotations or your thought process. I believe these questions will help you fire up your learning speed much more. 🔥 🔥

I strongly recommend going through each comments one by one and answering them all, since those questions were mostly made by me, asking myself when I first encountered Spring Boot.

If you have any questions, feel free to reach out via comment or DM.
Happy Hacking! 🌋 🌋 🌋

README.md
san-ghun: Thanks! I will keep this way!

src/main/kotlin/ecommerce/api/ProductController.kt
san-ghun: What I've noticed are those, `@Controller` can return 'view' with html file as UI but `@RestController` not. I think it is related with `@ResponseBody` annotation, because I tried to response with data through `@Controller` at first using `@ResponseBody`, but after using `@RestController` I don't need the `@ResponseBody` notation.

src/main/kotlin/ecommerce/api/ProductController.kt
san-ghun: Thanks for wrapping up, it make my mind clear and expend!
It seems correlated with 'Response Body' but has inner method to build the complete response programmatically, using `.build()`. 
I guess the alternative way is to make a response by adding custom header and body to build a response.

src/main/kotlin/ecommerce/api/ProductController.kt
san-ghun: Yeah, you're right. I was studying the difference between those http response status. 
Especially, I was not sure about the difference between 204 no content vs. 404 not found.
I now know that 2xx are the response for success and 4xx for client error. For more, 3xx is for redirection and 5xx for server error. Still figuring out 1xx codes but I feel much better after knowing the structure of these codes.

src/main/kotlin/ecommerce/api/Product.kt
san-ghun: I didn't recognized the term `entity` during the project, before PR. However, after the PR, I  noticed that there are terms related with `entity` such as, `DTO` and `DAO`. These terms are quite confusing for me right now.

My thought about `Entity` is that, it is a mapper or mapping object that correspond with DB table, so the application can manipulate and use the mapped row data in application. With this assumption, I think `Entity` can have logic to manipulate or process the column(?) data of the row. 
What if the product schema in DB change, `Product` domain doesn't have to deal with its constructor but with `toEntity()` method.
For now, this is my answer.

src/main/kotlin/ecommerce/api/ProductService.kt
san-ghun: I thought that `"""..."""` is a convention for SQL query, lol. My bad.
Now I know that it is multi-line string, and I can just use `"..."` single pair of double quote for single-line SQL query.

src/main/kotlin/ecommerce/api/ProductService.kt
san-ghun: Yes, it is better. Thanks!

src/main/kotlin/ecommerce/api/ProductService.kt
san-ghun: During the test code writing, I faced that if the queried data doesn't exist, the `queryForObject()` throws `DataAccessException`. What I expected to return in the `findById()` was `Product?`, not some exception. So, I used try-catch to make the function to work as I wanted. 

The `println` is for... kind of logging? I don't want the exception just pass and ignored, but rather left somehow. However, I still not exactly sure how to deal with it. 

What would be the better way to treat like this case?

src/main/kotlin/ecommerce/api/ProductService.kt
san-ghun: Yes, agree. It totally make sense and make code more readable. Thanks!

src/main/resources/application.properties
san-ghun: line 2-4
- Enable the H2 Console to use to view and inspect the current state of the configured H2 in-memory database.

line 5 and 6
- Not exactly know what those are. I wrote these two lines when I was configuring to use database by reading down some articles.

line 7-9
- To initialize the database to work by applying schema for table and insert sample data, whenever running the  application during the development. 
- line 7 is being commented out, because H2 in-memory database has default to init mode always.

src/main/kotlin/ecommerce/api/ProductController.kt
san-ghun: Interesting. Kotlin `Unit` corresponds to the `void` in Java. Since Spring Boot is a Java project, and I suppose that my project is going to be processed through JVM, so I might be down to `ResponseEntity<Void>`. Hope Spring Boot loves my thought 😂

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
san-ghun: I suppose it is to keep consistency of unit test. Each unit test shall be independent to each other, so there would be better to have new initiated table for `BeforeEach` tests. Thanks for presenting this subject up!

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
san-ghun: Both the annotation and the keyword were necessary for running the test. Since the test is relying on the JDBC with Spring Framework, the `JdbcTemplate` shall be managed by the Spring Framework. `@Autowired` annotation is needed to let the framework to manage the `JdbcTemplate`, and to initialize the `setup()` `@BeforeEach` test, `lateinit` were used to do so.

It seems a bit nasty to use annotations like this, but I suppose that for the combination of Java, Kotlin, with Spring Boot made this kind of thing happen. Maybe one day, when I become a super developer, I might PR for suggestion to change this unpleasant combination. LOL

src/main/kotlin/ecommerce/api/ProductService.kt
san-ghun: > `@Service` annotation seems new. What does this do, and what made you choose this annotation?

I used this annotation because other peer introduced this to me that the 'Service' layer is responsible to handle logic of application. I had thought that because `ProductService` was dealing with logic and data manipulation, it would be better to use this kind of "Bean" to make the application works better.

san-ghun: > I guess this question can be followed by your question: **Q. Since the current ProductService is providing direct interaction with database, using the Data Access Object (DAO). I think it might better to change it into ProductRepository.**
> 
> But before we talk about rather `ProductService` is a great name or not, it'll be great for us to go over what `service`, `DAO`, and `Repository` are. What are these, and what made you think that `ProductRepository` could be a better fit?

I suppose there are certain kind of pattern in Spring Boot, using Controller / Service / Repository / etc. A controller is responsible to deal with request and response. A service include logic to calculate and process data. Repository or Store(?) are dealing with Database.

I am still figuring out what `DAO` is. My current understanding is that, `DAO` is a pattern to use certain object as a buffer object to read and write data between application and data source. In conclusion, it seems like `Repository` in Spring Boot is an implementation of `DAO` pattern for Spring Framework.

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
san-ghun: Since the DB is an application or an environment external to our application, it is important to keep the state of shape of data and table (schema?). And each unit test is a independent test, so `@BeforeEach` test case there should be initiation of environment, in this case it is table in database. Then the test is reliable and predictable.

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
san-ghun: Frankly to say, I use `@JdbcTest` because one of provided learning material used this annotation in testing. There were two option from learning materials, one was with `@JdbcTest` and another was `@SpringBootTest` with `@AutoConfigureTestDatabase`. 
Both annotations were unknown to me, however reading down the documentation of two, I found `@JdbcTest` annotation is more readable to me. The other annotations were to broad for me. And, since I was using `JDBC` I choose to use `@JdbcTest` to run the test that needs database relation.

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
san-ghun: > I think there needs to be more than just `HttpStatus` assertion when it comes to testing api results, especially when it's related to `update`. However, the current update api doesn't allow us to check this with return value, since it's `ResponseEntity<Void>`. How could we check this?

I would refactor update logic to return `ResponseEntity<Product>` and add the product into the body of `200 OK`.

```kotlin
    @PutMapping("/products/{id}")
    fun updateProduct(
        @PathVariable id: Long,
        @RequestBody newProduct: Product,
    ): ResponseEntity<Product> {
        val result = productService.update(id, newProduct)   // refactor `.update()` to return number of affected row
        if (result == 1) {
            val product = productService.findById(id) ?: return ResponseEntity.notFound().build()
            return ResponseEntity.ok(product)
        }
        return ResponseEntity.notFound().build()
    }
```

san-ghun: > Interesting topic: If I attempt to update a product but the values are identical to the current ones, is it appropriate for the server to respond with `200 OK`? It'll be fun to talk about his with fellow crews 🙂
> 
> Also, I would like to know what you learned about: **Q. Explain the purpose and differences between HTTP methods (GET, POST, PUT, DELETE).** which you wrote in the PR message.

In general I would go with `200 OK` for the situation, because the application completed the request. However, if the purpose of application is protective and conservative to public, I would take alternative, like `204 No Content`.
Or, if those kind of behavior, like keep updating an identical value over and over, is judged as hostile activity, then it better be treated more seriously. 
But I have no idea how to deal with it in server-side. Maybe blocking the IP or rate-limiting?

---

- GET - to retrieve data, fetching a list or a single row(?)
- POST - to create new data, submitting a new form or register a row
- PUT - to replace existing data, editing or updating a FULL resource(?)
- DELETE - to delete existing data, removing a row

I found out there is another `PATCH` for partial update, but still figuring out how to use it.

san-ghun:
Hey PK!

Thanks for your thoughtful review and curated questions! From reviews that covering design and structure of the framework to coding conventions, I learn so much from your thoughtful review. Huge admiration to your warm heart and thoughtful brain!

# Main Changes

- In 'ProductController', refactor `createProduct()` and `updateProduct()` to return `ResponseEntity<Product>` rather than `ResponseEntity<Void>`
- Change naming for 'ProductService' into 'ProductRepository'

# Additional Note

With your curated questions in review comments, I tried to answer your topics with my own words in replying comments. A long the way, I learn so much about HTTP and Sprint Framework!

### ref) Questions

- Difference between `@Controller` and `@RestController`
- What is the role of `ResponseEntity`?
- What are the HTTP status codes?
- What is the `Entity`?
- `DTO` vs `Entity` vs `DAO`
- Difference between `Void` and `Unit` in Java and Kotlin
- What are the `@Service` and `@Repository`?
- What is the `@Autowired`? 
- What is the "Bean" in Spring Framework?
- Why and what are the usage for `@BeforeEach` in testing?
- What is the `JDBC` and why use `@JdbcTest` for testing it?
- What is the proper server response on certain situation?

src/main/kotlin/ecommerce/api/ProductController.kt
pkeugine: Yes, you're prediction is correct!
I actually recommend looking at the implementation of `@RestController` annotation.
It's quite simple and straight forward 😄

src/main/kotlin/ecommerce/api/ProductController.kt
pkeugine: Yeah, `ResponseEntity` feels like a tool that makes it easy for us to do that by (like you said,) building the response with inner methods. 👍

src/main/kotlin/ecommerce/api/Product.kt
pkeugine: Nice!
Yes, you're explanation about entity looks great.
And by looking at `Product`, for now, it feels like `Product` class is both a domain and an entity.
I asked this question because it's good to know what these terms mean when we put it in our projects.
You don't have to think too much about them though, if the terms sound confusing, since I felt like that too 😄 
It will come smoothly as time goes on.

src/main/kotlin/ecommerce/api/ProductService.kt
pkeugine: I agree on the point that instead of exception getting thrown, you wanted `null` to get passed since you want the consumer for this logic (in this case, it would be `ProductController`) to take care of this situation. That's actually very valid in practice, and it's not wrong :)

It's great that you tried logging when an exception gets thrown!
Usually, instead of `println`, we would use logging libraries to do it, but this seems fine for now.

I think it's okay to keep it like this for now! Nice work 👍

src/main/resources/application.properties
pkeugine: It would be great to also look up the ones you're not sure about,
and remove them if we're going to keep them commented out.

src/main/kotlin/ecommerce/api/ProductService.kt
pkeugine: Great. There's just soo many terms right? 😅 
I think you've done a great job dividing layers 👍 
Like you said, since service layer is more about `including logic to calculate and process data`,
the current `ProductService` class feels much more like `ProductRepository` or `ProductDao` to me.
The class is all about fetching or updating product data in DB.
How do you think?

---

update : I see you've changed it! 😄 👍

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
pkeugine: Perfect! 💯

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
pkeugine: The day will come, my friend. 😄

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
pkeugine: Reading the documentation 💯

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
pkeugine: I think thinking about `200 OK` is good enough for now!
It's very interesting to hear your thoughts though, since `updating an identical value over and over` will be one of the thinking points about update api in the future. Those kind of activities are the reasons we'll be caring about `idempotancy`, but it's not that important for now. I brought this up to say that your thoughts are valid!

pkeugine (approved):
Hello San, and well done answering the questions and refactoring the project!
I realized that I left a lot of questions while reviewing the PR 😅 , I hope it helped you get more close to Spring Boot.

I've added some responses to your answers, and I left an interesting topic in one of them. However, like I said in that comment, that topic is just for `there's things like this too!`, so you don't have to dive too deep into it.

I'll merge this PR, so if you have any questions, feel free to DM me anytime.
See you in the next PR, and until then,
Happy Hacking! 🔥 🔥 🔥

PR #8, opened 2025-07-17
[Spring-eCommerce-Product - Step1] Submit Vito's mission

VittorioDeMarzi:
# Pull Request: Step 1 – Product API

Hi 🙂  
Hope you're doing well!  
This PR wraps up the work for **Step 1-1 to Step 1-3** of the Product API mission.  
It covers a complete CRUD REST API for managing products — starting with in-memory storage and moving on to a proper H2 database using Spring JDBC.

---

### Step 1-1 – REST API (in-memory)

- Defined the `Product` data class
- Set up `ProductController` with:
  - `GET /products/{id}`
  - `GET /products`
  - `POST /products`
  - `PUT /products`
  - `DELETE /products/{id}`
- Used a simple `HashSet<Product>` for temporary storage

### Step 1-2 – Admin UI with Thymeleaf

- Hooked up Thymeleaf in the controller
- Created a `table.html` to display the products
- Images are rendered from URLs, no file upload involved

### Step 1-3 – Switch to H2 database

- Configured H2 in-memory database
- Enabled the console at `/h2-console`
- Created a `ProductRepository` using `JdbcTemplate`
- Added `schema.sql` and `data.sql` for DB setup
- Replaced the Kotlin collection logic with proper DB queries

---

## Quick Notes

I don’t have big open questions at this point, but I’d still be curious to hear your thoughts on:

- When using `JdbcTemplate`, is there a clean way to return the newly created or updated object straight from the controller?
- Am I missing any useful Kotlin best practices that you'd suggest applying here?

Thanks a lot for reviewing! Let me know if you see anything worth changing or improving.


README.md
```diff
@@ -1 +1,62 @@
-# spring-ecommerce-product
\ No newline at end of file
+# spring-ecommerce-product
+
+## Functional Requirements Step 1-1
+
+Implement a simple HTTP API that allows users to retrieve, add, update, and delete products.
```
da-nyee, line 5: Well documented 👍

build.gradle.kts
```diff
@@ -23,12 +23,14 @@ dependencies {
...
     implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("nz.net.ultraq.thymeleaf:thymeleaf-layout-dialect:3.2.0")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     runtimeOnly("com.h2database:h2")
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("io.rest-assured:rest-assured:5.3.1")
 }
```
da-nyee, line 34: I'm just wondering — do you know the difference between `implementation`, `api`, `runtimeOnly`, and `compileOnly`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+
+@Controller
```
da-nyee: What is the difference between `@Controller` vs `@RestController`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,64 @@
...
+@RequestMapping("/products")
+class ProductController {
+    @Autowired
+    private lateinit var productRepository: ProductRepository
```
da-nyee: I noticed you're using `Field Injection` for DI 👀
Have you heard of `Constructor Injection` and `Setter Injection` as well?
How do they differ, and which one is generally recommended?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    @PostMapping(consumes = ["application/json"])
+    @ResponseBody
+    fun createProduct(
+        @RequestBody product: Product,
+    ): ResponseEntity<Product> {
```
da-nyee: Is it okay to use a model object as either a request or a response?

da-nyee: There might be a better object for this 🤔

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun getProducts(model: Model): String {
+        model.addAttribute("products", productRepository.getAll())
+        return "table"
+    }
+
+    @GetMapping("/{id}")
+    fun getProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> {
+        val product = productRepository.findById(id) ?: throw NoSuchElementException()
```
da-nyee: image

Seems like the exception isn't imported — could you check it?

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.model
+
+data class Product(
```
da-nyee: I'm just curious — is `data class` a good approach for model objects?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,51 @@
...
+            Product(
+                rs.getLong("id"),
+                rs.getString("name"),
+                rs.getDouble("price"),
+                rs.getString("image_url"),
+            )
+        }
+
+    fun findById(id: Long): Product? {
+        return jdbcTemplate.queryForObject("select id, name, price, image_url from products where id = ?", productRowMapper, id)
```
da-nyee: > When using `JdbcTemplate`, is there a clean way to return the newly created or updated object straight from the controller?

As I mentioned in a comment on `ProductController`, there might be a more appropriate object for transferring data 👀

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import io.restassured.RestAssured
+import io.restassured.http.ContentType
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
da-nyee, line 17: Could you compare the `classMode` options?

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,100 @@
...
+                .then()
+                .extract()
+                .response()
+
+        assertThat(response.statusCode).isEqualTo(HttpStatus.OK.value())
+        assertThat(response.body.jsonPath().getList<Product>("")).hasSize(3)
+    }
+
+    @Test
+    fun getProduct() {
```
da-nyee, line 115: image

I see some tests have failed — could you take a look?

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
```diff
@@ -0,0 +1,86 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.JdbcTest
+import org.springframework.jdbc.core.JdbcTemplate
+
+@JdbcTest
+class ProductRepositoryTest {
+    private lateinit var productRepository: ProductRepository
```
da-nyee, line 13: What's the difference between `lateinit` and `by lazy`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,64 @@
...
+        return ResponseEntity.ok(product)
+    }
+
+    @PutMapping
```
da-nyee: Are you familiar with `@PatchMapping`?
Could you explain the difference between `@PutMapping` and `@PatchMapping`?

da-nyee: And what about `@PostMapping`?

da-nyee (changes requested):
Hi Vito! 🙌
Thanks for waiting for the feedback 🙇‍♀️

You've done well 👍
But there's some room for improvement, so I've left a few comments.
If anything's unclear, feel free to let me know!

src/main/kotlin/ecommerce/controller/ProductController.kt
VittorioDeMarzi: The @Controller annotation is used in Spring MVC when you're building a traditional web application that returns HTML pages to the browser.
The @RestController is used to build RESTful web services. It’s a convenience annotation that combines @Controller and @ResponseBody.

I've now separated the concerns between the API and the frontend view implementing two controllers. Hope this is a good practice 😄

build.gradle.kts
VittorioDeMarzi: Thanks for asking, I hadn't questioned yet the differences between them. These dependencies were already part of the initial project setup, we haven’t added any custom ones yet.

To quickly recap the difference:
- implementation: adds the dependency for compile and runtime, but hides it from consumers (good for encapsulation)
- api: same as implementation, but exposes the dependency to modules that depend on this one
- runtimeOnly: used only at runtime, not available at compile time (e.g. for in-memory databases like H2)
- compileOnly: available at compile time but excluded from the runtime classpath (e.g. annotations provided by the environment)

I'll keep in mind this for the future 🙂 and let me know please if you'd suggest any adjustments!

src/main/kotlin/ecommerce/controller/ProductController.kt
VittorioDeMarzi: Great catch 👀

- Constructor Injection: the dependency is provided via the class constructor. It's immutable and makes the dependency mandatory, which helps with testability and readability.
- Setter Injection: the dependency is set through a method or property, which allows it to be optional or replaceable.
- Field Injection: quick and concise, but not ideal for testing or immutability.

From what I’ve read and seen, constructor injection is generally the most recommended, especially in Kotlin, since, @autowired is not needed, it works well with val and avoids reflection.

Happy to refactor it 🙃

src/main/kotlin/ecommerce/controller/ProductController.kt
VittorioDeMarzi: At first I was using the domain model as both request and response, but then I refactored and introduced dedicated DTOs (ProductRequest and ProductResponse) to better separate concerns and avoid exposing internal structure.

Let me know if the current approach can be further improved 🙏

src/main/kotlin/ecommerce/controller/ProductController.kt
VittorioDeMarzi: For now I changed it in 
val product = productRepository.findById(id) ?: return ResponseEntity.notFound().build()

but I will have to manage Exceptions in step2 implementing an ExceptionHandler class.

src/main/kotlin/ecommerce/model/Product.kt
VittorioDeMarzi: Good question! Initially I used a data class for Product out of convenience, but I realized it's not ideal for model objects, especially when they have business logic or mutable state.
In fact, I’ve now refactored the domain model and removed data class in favor of a regular class to better reflect behavior and maintain encapsulation.

Thanks for making me reflect on this! 🙏

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
VittorioDeMarzi: @DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD) tells Spring to reset the application context before each test method. This ensures full isolation between tests — any changes to the database or beans in one test won't affect the others.

Here's a quick comparison of the available options:
- BEFORE_EACH_TEST_METHOD: Reloads the context before each test method → safest for test isolation, but slower.
- AFTER_EACH_TEST_METHOD: Reloads the context after each test method → same isolation level, just delayed cleanup.
- BEFORE_CLASS: Reloads the context once before the test class → faster, but shared state between tests.
- AFTER_CLASS: Reloads the context once after the test class → similar to above, but delays reset until the end.

Is there anything I should pay attention to?

src/main/kotlin/ecommerce/controller/ProductController.kt
VittorioDeMarzi: @PostMapping → Create a new resource
- Create new entries in a database (e.g., a new product)

@PutMapping → Replace the entire resource
- Full update of an existing object
With PUT, all fields must be present in the request, even unchanged ones.

@PatchMapping → Partially update a resource
- Update just part of the resource
Changing only specific fields (e.g., just the price)

I will have to implement @Patch Mapping for the second step :)
Is there anything I should be aware of at this point?

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
VittorioDeMarzi: ## Lateinit
Lateinit is a keyword used to declare a non-nullable variable that will be initialized later. You promise Kotlin that you'll initialize the variable before you use it, and Kotlin trusts you. However, if you try to access the variable before initializing it, you'll get a crash at runtime.
  - It can only be used with var (mutable variables).
  - It cannot be used with primitive data types (e.g., Int, Double).
  - “I promise I’ll initialize it before use!”
  
```
@JdbcTest
class ProductRepositoryTest {
    private lateinit var productRepository: ProductRepository

    @Autowired
    private lateinit var jdbcTemplate: JdbcTemplate
```
## Lazy
Lazy is a delegate that initializes a value only when it’s accessed for the first time. It’s used for variables that are read-only (val) and when initialization is computationally expensive or unnecessary until the value is used.
  - It can only be used with val (immutable variables).
  - Initialization happens lazily and is thread-safe by default.
  - “I'll only initialize it if and when it's needed.”
  
```
@JdbcTest
class ProductRepositoryTest {
    private val productRepository: ProductRepository by lazy {
        ProductRepository(jdbcTemplate)
    }

    @Autowired
    private lateinit var jdbcTemplate: JdbcTemplate
```
  
In my case they are both working. Which should I prefer?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
VittorioDeMarzi: Good point. As I said before I’ve now separated the domain model from the API by introducing ProductRequest and ProductResponse DTOs but I am creating the in the controller.

Should give this responsibility to the controller?
Should I use SimpleJdbcInsert?

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
VittorioDeMarzi: @DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD) tells Spring to reset the application context before each test method. This ensures full isolation between tests, any changes to the database or beans in one test won't affect the others.

Here's a quick comparison of the available options:
- BEFORE_EACH_TEST_METHOD: Reloads the context before each test method → safest for test isolation, but slower.
- AFTER_EACH_TEST_METHOD: Reloads the context after each test method → same isolation level, just delayed cleanup.
- BEFORE_CLASS: Reloads the context once before the test class → faster, but shared state between tests.
- AFTER_CLASS: Reloads the context once after the test class → similar to above, but delays reset until the end.

I used BEFORE_EACH_TEST_METHOD here to make sure each test starts from a clean state

VittorioDeMarzi:
Hi Da-nyee! 😊
Thanks so much for all the thoughtful and insightful feedback, it really helped me reflect on several design decisions 🙏

After thinking through your questions, I made a few changes in the codebase, I hope I’ve addressed them correctly! I'm still open to further feedback if anything could be improved.

For now, I’ve pushed the updates based on your comments. In the meantime, I’ve also started working on introducing a ProductService class with its interface, and I'm also setting up an ExceptionHandler class, I’ll continue with the implementation later this morning and push the next changes once ready 🚀

Thanks again for guiding this learning process so clearly. Looking forward to your future suggestions! 🙌

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,155 @@
...
+
+        assertThat(response.statusCode).isEqualTo(HttpStatus.OK.value())
+        assertThat(response.`as`(ProductResponse::class.java).id).isEqualTo(1L)
+        assertThat(response.`as`(ProductResponse::class.java).name).isEqualTo("cafe")
+        assertThat(response.`as`(ProductResponse::class.java).price).isEqualTo(39.00)
+        assertThat(response.`as`(ProductResponse::class.java).imageUrl).isEqualTo("www.test")
+    }
+
+    @Test
+    fun deleteProduct() {
```
da-nyee, line 133: image

Could you take a look at this test?

src/main/kotlin/ecommerce/repository/ProductResponse.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.repository
+
+data class ProductResponse(
```
da-nyee: It might be a good idea to place this response in the `dto` package 👀

src/main/kotlin/ecommerce/controller/ProductController.kt
da-nyee: Looking forward to seeing the exception handler 😏

src/main/kotlin/ecommerce/repository/ProductRepository.kt
da-nyee: > Should give this responsibility to the controller?

If I understood correctly, it means you've put business logic in the controller, right?
I think it's fine for now, but you might consider introducing a `service` layer.
Feel free to add it in the next step if you'd like.

And if I misunderstood anything, please let me know 🙇‍♀️

<br/>

> Should I use SimpleJdbcInsert?

Well, I guess that's okay — you don't need to use it.

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
da-nyee: > In my case they are both working. Which should I prefer?

Since you're using DI, `lateinit` is usually the better choice — especially when combined with `@BeforeEach`.
It keeps the initialization simple and explicit.

src/main/kotlin/ecommerce/controller/ProductController.kt
da-nyee: > Is there anything I should be aware of at this point?

You've understood the differences well, so there's nothing in particular to worry about for now 🤔

da-nyee (changes requested):
Hi Vito! 🙌
You're doing great so far 👍

There's just one test failure — could you take another look?
Once it passes, I'll go ahead and merge this PR!

I've also left a few comments, so please check them out when you get a chance 🙇‍♀️

VittorioDeMarzi:
Hi @da-nyee! 😊
Thank you so much for your feedback and encouragement, I really appreciate the time you took to go through everything so thoroughly! 🙇‍♂️

I’ve:
- Created a Service Layer to encapsulate business logic:
- Defined a ProductService interface with methods like findById, createProduct, updateProduct, getAll, and delete.
- Provided a concrete implementation ProductServiceImpl that delegates to the ProductRepository.
- Implemented centralized exception handling using @RestControllerAdvice:
- ProductNotFoundException → 404 Not Found
- ProductCreationException and ProductUpdateException → 500 Internal Server Error
- Generic RuntimeException → 400 Bad Request
- Added a consistent error response format via a custom ErrorMessageModel.

Let me know if there’s anything else I should improve, I’m learning a lot through this! 🙏

src/main/kotlin/ecommerce/service/impl/ProductServiceImpl.kt
```diff
@@ -0,0 +1,52 @@
...
+import ecommerce.exception.ProductCreationException
+import ecommerce.exception.ProductNotFoundException
+import ecommerce.exception.ProductUpdateException
+import ecommerce.model.toDto
+import ecommerce.repository.ProductRepository
+import ecommerce.service.ProductService
+import org.springframework.stereotype.Service
+
+@Service
+class ProductServiceImpl(private val productRepository: ProductRepository) : ProductService {
```
da-nyee, line 14: Nice work 👍
But I think the interface might be a bit redundant for now, since it currently has only one implementation 👀

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,49 @@
+package ecommerce.exception
+
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.RestControllerAdvice
+
+@RestControllerAdvice
+class GlobalExceptionHandler {
```
da-nyee, line 9: 👍

src/main/kotlin/ecommerce/exception/ErrorMessageModel.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.exception
+
+class ErrorMessageModel(
```
da-nyee, line 3: 👍

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    @ExceptionHandler(RuntimeException::class)
+    fun handleRunTime(ex: RuntimeException): ResponseEntity<ErrorMessageModel> {
+        val errorMessage =
+            ErrorMessageModel(
+                HttpStatus.BAD_REQUEST.value(),
+                ex.message,
+            )
+        return ResponseEntity(errorMessage, HttpStatus.BAD_REQUEST)
```
da-nyee, line 27: Is there any particular reason you've mapped general runtime exceptions to 400(Bad Request)?

da-nyee (approved):
Hi Vito! 🙌

All the tests are passing — thanks for applying the feedback ☺️
I've also gone through the latest changes — awesome 👍

Everything looks good to move on, so let's wrap up this PR!
See you in the next step 🤸‍♀️

PR #9, opened 2025-07-17
[spring-ecommerce-product - Step 1-1 to 1-3] Submit Farhana's mission

farhanaahmed:
eCommerce: Products - Introduction to Spring
This PR implements the full scope of the eCommerce: Products - Introduction to Spring mission.
✅ Step 1-1: Implement Product CRUD with In-Memory Kotlin Collections
Added `Product` data class
Built CRUD operations using a mutable list in memory
Implemented routes and views
Tested functionality locally via browser
✅ Step 1-2: Implement an admin interface
Created basic HTML pages for admin product management
Setup basic controller, service, and repository layers
✅ Step 1-3: Replace In-Memory with JDBC + H2
Removed in-memory collection logic
Introduced Spring's JdbcTemplate to perform database operations
Added schema.sql under resources/ to initialize DB
Enabled H2 console for debugging
Added integration tests for ProductRepository


.gitignore
```diff
@@ -38,3 +38,6 @@ out/
 
 ### Mac OS ###
 .DS_Store
+
+productdb.mv.db
+productdb.trace.db
```
Hyeon9mak, line 43: git ignore managing 👍 💯 
https://medium.com/p/b42d14dccc93 😉

README.md
```diff
@@ -1 +1,54 @@
...
+
+## Integrate DB in Project `Product` -> `Step 1-3`
+- [x] install db dependency
+- [x] create schema.sql
+- [x] create data.sql
+- [x] add rules in application.properties
+- [x] create table
+- [x] Convert Controller for db usage
+- [x] Create Product Repository to handle DB logic
+- [ ] Implement Services to move business logic from Controller
```
Hyeon9mak: nice habit 👍 
the last thing... 😉

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice
+class GlobalExceptionHandler {
+    @ExceptionHandler(NotFoundException::class)
+    fun handleNotFoundException(e: NotFoundException): ResponseEntity<Void> {
+        println("NotFoundException occurred: " + e.message)
+        return ResponseEntity.notFound().build()
+    }
+}
```
Hyeon9mak, line 15: Global exception handler! 👍 💯 
What's different between `@ControllerAdvice` & `@RestControllerAdvice` ?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import java.net.URI
+
+@Controller
+@RequestMapping("/products")
+class ProductController {
```
Hyeon9mak: `@RequestMapping`! 👍 
What's different between `@Controller` & `@RestController` ?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,70 @@
...
+    @Autowired
+    private lateinit var jdbcTemplate: JdbcTemplate
+
+    @PostMapping("")
```
Hyeon9mak: used `@Autowired` is anti-pattern in SpringBoot now.
How about use constructor-injection pattern?
https://www.baeldung.com/constructor-injection-in-spring

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @PutMapping("/{id}")
+    fun update(
+        @PathVariable("id") id: Long,
+        @RequestBody newProduct: Product,
+    ): ResponseEntity<Void> {
+        ProductRepository(jdbcTemplate).update(id, newProduct)
+        return ResponseEntity.ok().build()
+    }
```
Hyeon9mak: What's different between HTTP method PUT & PATCH?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @DeleteMapping("/{id}")
+    fun delete(
+        @PathVariable("id") id: Long,
+    ): ResponseEntity<Void> {
+        ProductRepository(jdbcTemplate).deleteById(id)
+        return ResponseEntity.noContent().build()
+    }
```
Hyeon9mak: 204 status code. niiice 💯

src/main/kotlin/ecommerce/exception/NotFoundException.kt
```diff
@@ -0,0 +1,3 @@
+package ecommerce.exception
+
+class NotFoundException(message: String) : RuntimeException(message)
```
Hyeon9mak, line 3: Why did u generate this custom exception?
(just ask question. feel free 😄 )

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,23 @@
...
+    val id: Long? = null,
+    var name: String = "",
+    var price: Double = 0.0,
+    var imageUrl: String = "",
+) {
+    fun update(newProduct: Product) {
+        this.name = newProduct.name
+        this.price = newProduct.price
+        this.imageUrl = newProduct.imageUrl
+    }
```
Hyeon9mak: In the context of entities, equality is determined based on the ID.

For example, if you have

- id = 1, name = "red apple"

On the other hand

```kotlin

        other as Product

        return id == other.id
    }

        return id?.hashCode() ?: 0
    }
```

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,53 @@
...
+        product: Product,
+    ) {
+        val sql = "update products set product_name = ?, price = ?, image_url = ? where id = ?"
+        jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, id)
+    }
+
+    fun deleteById(id: Long) {
+        val sql = "delete from products where id = ?"
+        jdbcTemplate.update(sql, id)
+    }
```
Hyeon9mak: how do we know update or delete failed?

src/main/resources/schema.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE IF NOT EXISTS products (
+    id INT PRIMARY KEY AUTO_INCREMENT,
+    product_name VARCHAR(255) NOT NULL,
+    price DOUBLE CHECK (price >= 0),
+    image_url VARCHAR(255)
+    );
```
Hyeon9mak, line 6: name & price & image have limits.

I think adding validation logic to the Product, might make the code more robust and easier to test.
U think so too, right? 😄

Hyeon9mak (changes requested):
Hello Farhana! Welcome to Spring mission 🥳 

Implemented everything very nice!
I think u already know Spring well.
Just left a few minor comments. 😄 

.gitignore
farhanaahmed: The article was about the “no newline at end of file” warning on GitHub. I found it really helpful. I’ve always wondered why that warning appears, and now I finally understand the reason behind it. It feels so good to know.😄 Thanks a lot for sharing this.

README.md
farhanaahmed: Thanks for the feedback. 😊 I’ve removed the last feature for now to keep things clean. I’ll work on that as part of step 2.

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
farhanaahmed: Thank you! 😊
The main difference is that `@RestControllerAdvice` is a specialization of `@ControllerAdvice` that automatically adds `@ResponseBody` to all methods. So the methods return JSON (or other response bodies) directly, which is useful for REST APIs.
In contrast, `@ControllerAdvice` is typically used in web apps with views (e.g. Thymeleaf), where exceptions might redirect to a HTML page instead of returning raw JSON.

src/main/kotlin/ecommerce/controller/ProductController.kt
farhanaahmed: `@RestController` is a shortcut for `@Controller + @ResponseBody`, so it's mainly used for building pure REST APIs and when the responses need to be serialized automatically (like to JSON).
In my case, I used `@Controller` because I'm handling both: rendering views with Thymeleaf and returning some data as JSON. So, `@Controller` gives me more flexibility to mix both approaches.

src/main/kotlin/ecommerce/controller/ProductController.kt
farhanaahmed: Thank you for the feedback. I have removed `@Autowired` from the controller and replaced it with constructor injection as recommended. However, I’ve kept using `@Autowired` in the test files since it’s a common and convenient practice there.But please let me know if you have any better approach to do it. Thank you for also including the documentation link. 🙏

src/main/kotlin/ecommerce/controller/ProductController.kt
farhanaahmed: The main difference between **PUT** and **PATCH** is how they update resources:

- **PUT** expects the **entire resource representation** and replaces the existing resource with what is sent. If some fields are missing, they may be reset or cleared.

- **PATCH** is meant for **partial updates.** Here, the user sends only the fields that need to be changed, and the server applies those changes without affecting other fields. This is useful for making small changes without resending the whole object.

In my current implementation, I used **PUT** because I expect the client to send a complete `Product` object when updating. This simplifies the update logic since I replace the entire product data for that ID. It ensures consistency by always having a full resource state on the server.

If in the future I want to support partial updates, I could add a **PATCH** endpoint to handle cases where only some fields change. But for now, using PUT keeps the API simpler and aligns well with the way the update method is written.

Please let me know your thoughts on this topic. 🙏

src/main/kotlin/ecommerce/exception/NotFoundException.kt
farhanaahmed: Initially, I created the custom `NotFoundException` to handle specific error cases in a centralized way using a separate exception handler class. My goal was to keep the code clean and maintainable by separating error-handling logic. However, as the project progressed, I realized that this was the only exception I needed, and I haven’t encountered other exception cases yet. So I’m wondering whether it’s still worth keeping this custom exception or if it would be better to simplify the code and just use a built-in exception instead. I would like to know your thoughts about this. 🙏

src/main/kotlin/ecommerce/model/Product.kt
farhanaahmed: I understand your logic about determining equality based on the entity ID, and I have applied the recommended changes to my Product class by overriding equals and hashCode to use only the id. This ensures that products with the same ID are considered equal, even if other attributes change. Thank you for the guidance. 🙏

src/main/kotlin/ecommerce/repository/ProductRepository.kt
farhanaahmed: Thanks for pointing that out. 🙌

I've updated the update and delete methods to return the affected row count. This way, I can now verify whether the operation actually modified or deleted any record.

I also updated the CollectionProductStore accordingly to match the new method return type. 

Let me know if you have any further suggestions.

src/main/resources/schema.sql
farhanaahmed: Yes, I totally agree that adding validation logic in the Product class makes the code more robust and easier to test. I've added the necessary require checks to validate name, price, and image URL constraints.

I've also created a corresponding unit test class (ProductTest) to ensure that invalid values correctly throw exceptions and valid products are created as expected.

Thanks for the suggestion. 😊

I’m keeping the SQL constraints in the CREATE TABLE script as an extra safety layer to ensure data integrity from multiple fronts. Let me know if you'd recommend handling it differently!

farhanaahmed:
Hello, Gumak. 😄  Sorry for the delayed update. 

During class, Brie showed us how we can use an interface to implement both the in-memory collection and the database. I wanted to take some time to try that approach and update my code accordingly.

Could you please check if I’ve implemented it successfully?

I also tried to address your previous comments and went through the materials you shared, they were really helpful. Thank you so much for your time and valuable feedback!

I also have another small question about something. I’m seeing this warning message every time I run the application:

“Mockito is currently self-attaching to enable the inline-mock-maker. This will no longer work in future releases of the JDK. Please add Mockito as an agent to your build as described in Mockito's documentation.”

However, I haven’t explicitly used Mockito in my test files. Should I ignore this warning for now, or is there something I should do to address it? Looking forward to your advice. Thank you so much. 🙏

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
Hyeon9mak: 💯

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,73 @@
...
+
+@Controller
+@RequestMapping("/products")
+class ProductController(
+    @Qualifier("jdbcProductStore")
+    private val productStore: ProductStore,
+) {
+    @PostMapping("")
+    @ResponseBody
+    fun create(
```
Hyeon9mak, line 27: > However, I’ve kept using `@Autowired` in the test files since it’s a common and convenient practice there.

In certain cases like writing test code, I think it’s totally acceptable to use it.
If writing tests is more convenient, you’ll naturally run them more often and keep things in check more frequently.
Well done! 👍

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    @PutMapping("/{id}")
+    @ResponseBody
+    fun update(
+        @PathVariable("id") id: Long,
+        @RequestBody newProduct: Product,
+    ): ResponseEntity<Void> {
+        productStore.update(id, newProduct)
+        return ResponseEntity.ok().build()
+    }
```
Hyeon9mak, line 48: > The main difference between PUT and PATCH is how they update resources:.....

Ur answer is so perfect that I have nothing to add! 💯

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,73 @@
...
+    @PostMapping("")
+    @ResponseBody
+    fun create(
+        @RequestBody product: Product,
+    ): ResponseEntity<Void> {
+        val id = productStore.save(product)
+        return ResponseEntity.created(URI.create("/products/$id")).build()
+    }
+
+    @GetMapping("")
```
Hyeon9mak, line 34:
```suggestion
    @GetMapping
```
😉

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,73 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.ResponseBody
+import java.net.URI
+
+@Controller
+@RequestMapping("/products")
```
Hyeon9mak, line 20: > In my case, I used `@Controller` because I'm handling both: rendering views with Thymeleaf and returning some data as JSON. So, `@Controller` gives me more flexibility to mix both approaches.

💯 
also u can split two controllers
for JSON & Thymeleaf.

src/main/kotlin/ecommerce/exception/NotFoundException.kt
Hyeon9mak: It's best to create a custom exception only when you really need to clearly distinguish a specific case. 
Otherwise, I think it's better to use the standard exceptions provided by default, since it makes it easier for other developers to quickly understand the reason behind the exception.

src/test/kotlin/ecommerce/model/ProductTest.kt
```diff
@@ -0,0 +1,56 @@
...
+            Product(
+                id = 1,
+                name = "Apple",
+                price = 1.99,
+                imageUrl = "http://image.com/apple.jpg",
+            )
+        assertEquals("Apple", product.name)
+    }
+
+    @Test
```
Hyeon9mak, line 19: 😉 👍

src/main/kotlin/ecommerce/JdbcProductStore.kt
```diff
@@ -0,0 +1,55 @@
...
+        product: Product,
+    ): Int {
+        val sql = "update products set product_name = ?, price = ?, image_url = ? where id = ?"
+        return jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, id)
+    }
+
+    override fun delete(id: Long): Int {
+        val sql = "delete from products where id = ?"
+        return jdbcTemplate.update(sql, id)
+    }
```
Hyeon9mak, line 54: Returning the number of affected rows might be a concept more relevant to the persistence(RDBMS) layer.
From the perspective of the ecommerce domain, it's not necessarily an important concern.

That's why it could be a good idea to throw an exception or handle it in some way if the number of affected rows during an update or delete isn't exactly one.

src/main/resources/schema.sql
Hyeon9mak: Adding such constraints to ensure data integrity is a good approach — well done.

In real-world applications, however, there are often cases where you need to manually access and correct data directly in the database. These constraints can sometimes make such maintenance difficult.

While applications can scale horizontally to avoid single points of failure during high traffic, the database tends to become a bottleneck more easily. That's why we often prefer to enforce such validations at the application code level to minimize the load on the DB.

Still, I think this was a great experiment — good job trying it out! 👍

src/main/kotlin/ecommerce/CollectionProductStore.kt
```diff
@@ -0,0 +1,42 @@
...
+import ecommerce.model.Product
+import org.springframework.stereotype.Repository
+
+@Repository
+class CollectionProductStore(val products: MutableList<Product>) : ProductStore {
+    override fun countProducts(): Int {
+        return products.size
+    }
+
+    override fun findAll(): List<Product> {
```
Hyeon9mak, line 13: Since this is only used in tests, you can move it under the test package.
Also, since it doesn't use Spring's container for bean injection, you can remove the annotations as well.

```suggestion
class CollectionProductStore(
    private val products: MutableList<Product>,
) : ProductStore {
    override fun countProducts(): Int {
        return products.size
    }

    override fun findAll(): List<Product> {
```

src/main/kotlin/ecommerce/CollectionProductStore.kt
```diff
@@ -0,0 +1,42 @@
...
+        return products.toList()
+    }
+
+    override fun findById(id: Long): Product? {
+        return products.firstOrNull { it.id == id } ?: throw NotFoundException("Product with id $id not found")
+    }
+
```
Hyeon9mak, line 20:
```suggestion
    override fun findById(id: Long): Product {
        return products.firstOrNull { it.id == id } ?: throw NotFoundException("Product with id $id not found")
    }

```
😉 

```kotlin
override fun update(
        id: Long,
        product: Product,
    ): Int {
        val updatedProduct = findById(id)
        updatedProduct.name = product.name
        updatedProduct.price = product.price
        updatedProduct.imageUrl = product.imageUrl
        return 1
    }
```

Hyeon9mak (approved):
Hello Farhana! Nice work 👍 
Your answers show a high level of thought. Great job. 💯 

The reason for the Mockito-related error is that `org.springframework.boot:spring-boot-starter-test` includes a Mockito dependency, and with newer JDK versions, dynamic attachment will be restricted in the future.
This warning is suggesting that you prepare for that.

Since you're not using Mockito at the moment, you can either exclude it
or, if you do want to write tests using Mockito, add the appropriate configuration.

```
// build.gradle.kts
testImplementation("org.springframework.boot:spring-boot-starter-test") {
    exclude(group = "org.mockito")
}
```
```
// build.gradle.kts
tasks.withType<Test> {
    useJUnitPlatform()
    jvmArgs("-XX:+EnableDynamicAgentLoading")
}
```
or follow the [official docs](https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3)
image


I’ve left a few comments—please take your time to review them as you move on to the next step.
See you in the next step!

PR #10, opened 2025-07-17
[Spring-Ecommerce-Product Step 1] Submit arekiu's mission

arekiu:
# spring-ecommerce-product

Building a basic e-commerce using Kotlin and Spring Boot framework.

### Features step 1.1
- Implement a simple HTTP API that allows users to retrieve, add, update, and delete products.

### Features step 1.2
- Implement an admin interface that allows users to view, add, update, and delete products.

### Features step 1.3
- Store product information in H2 in-memory database and implement methods to interact with it.

## Implementation
- Create ProductController class for API routes.
- Create ProductService class to handle DataBase logic.
- Create ProductViewController to manage page routes.
- Introduce Templates for the views for Products, Edit and Create pages.
- Use JavaScript for frontend functionality fetching data.

## Reviewer notes

- It´s my first time working with Spring and I would like to know if the structure of the project makes sense.
- In the beginning we were giving data to the view through ProductViewController. But finally we changed to receive the data in the frontend using JavaScript to fetch the data. I would like to know what is better approach.
- We are using products directly from the database. Is it necessary to create a new list of objects to process them?

Thank you so much for your time for reviewing my code!


src/main/kotlin/ecommerce/ProductViewController.kt
```diff
@@ -0,0 +1,25 @@
+package ecommerce
```
da-nyee: > It´s my first time working with Spring and I would like to know if the structure of the project makes sense.

You did a good job 👍
I didn't know you were new to the Spring Framework 🌱

But I'd recommend making the structure a bit more specific — for example, by adding packages such as `controller` or `model`.

src/main/resources/templates/createProductForm.html
```diff
@@ -0,0 +1,91 @@
...
+            </label>
+        </fieldset>
+        <input
+                type="submit"
+                value="Create"
+        />
+    </form>
+</main>
+</body>
+<script>
```
da-nyee, line 64: > In the beginning we were giving data to the view through ProductViewController. But finally we changed to receive the data in the frontend using JavaScript to fetch the data. I would like to know what is better approach.

You're doing great — it's awesome that you can work with JS as well 👍

But how about handling the data in the controller instead, since we're focusing on practicing the Spring Framework?

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
...
+                db.queryForObject(
+                    """select id, name, price, imageUrl from product where id = ?""",
+                    productRowMapper,
+                    id,
+                )
+        } catch (exception: Exception) {
+            println(exception.message)
+            return null
+        }
+        return product
```
da-nyee: > We are using products directly from the database. Is it necessary to create a new list of objects to process them?

Yeah, you might consider introducing a new object called `DTO(Data Transfer Object)` 👀
What do you think about using it?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
```
da-nyee: What's the difference between `@Controller` and `@RestController`?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    @GetMapping("/products/{id}")
+    fun getProduct(
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> {
+        val product = productService.findById(id) ?: return ResponseEntity.notFound().build()
+        return ResponseEntity.ok(product)
+    }
+
+    @PutMapping("/products/{id}")
```
da-nyee: Have you heard of `@PatchMapping`?
What's the difference between `@PutMapping` and `@PatchMapping`?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+    fun updateProduct(
+        @PathVariable id: Long,
+        @RequestBody newProduct: Product,
+    ): ResponseEntity<Void> {
+        productService.findById(id) ?: return ResponseEntity.notFound().build()
+        productService.update(id, newProduct)
```
da-nyee: You might consider grouping these methods in the service 👀

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
+package ecommerce.api
+
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.stereotype.Service
+import java.sql.ResultSet
+
+@Service
+class ProductService(private val db: JdbcTemplate) {
```
da-nyee: I guess this class seems more like a repository.
Do you know the difference between a service and a repository?

src/main/kotlin/ecommerce/api/ProductService.kt
```diff
@@ -0,0 +1,67 @@
+package ecommerce.api
+
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.stereotype.Service
+import java.sql.ResultSet
+
+@Service
```
da-nyee: What's the difference between `@Component` and `@Service`?

src/main/resources/application.properties
```diff
@@ -1 +1,9 @@
 spring.application.name=spring-ecommerce
+spring.h2.console.enabled=true
+spring.h2.console.path=/h2-console
+spring.datasource.url=jdbc:h2:mem:database
+#spring.jpa.defer-datasource-initialization=true
+#spring.jpa.hibernate.ddl-auto=none
+#spring.sql.init.mode=always
```
da-nyee: I'd recommend removing the comments if they aren't needed.

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
```diff
@@ -0,0 +1,91 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.JdbcTest
+import org.springframework.http.HttpStatus
+import org.springframework.jdbc.core.JdbcTemplate
+
+@JdbcTest
+class ProductControllerTest {
```
da-nyee: Is the annotation appropriate for this test?

src/main/kotlin/ecommerce/api/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
+@RequestMapping("/api")
+class ProductController(private val productService: ProductService) {
```
da-nyee: You're using `Constructor Injection` for DI 👍 
Have you heard of `Field Injection` and `Setter Injection` as well? Do you know how they differ?

da-nyee (changes requested):
Hi Arekiu! 🙌
You're doing great 👍

I've answered your questions and left a few comments, as there's some room for improvement.
If you have any other questions, just let me know 🙇‍♀️

src/main/kotlin/ecommerce/api/ProductService.kt
arekiu: I´m taking time to understand it. So, Service and Repository are Components. Service is used for Business Logic and Repository for consulting the Data Base.
I´m working in something to improve this logic.

src/main/kotlin/ecommerce/api/ProductService.kt
arekiu: A Component could be something more generic, like a Class used for other utilities.

src/main/kotlin/ecommerce/api/ProductController.kt
arekiu: @Controller returns views (like the HTML pages).
@RestController is used to build APIs, returning data.

src/main/kotlin/ecommerce/api/ProductService.kt
arekiu: Finally, I created a @Repository to deal with DataBase and @Service is cleaner.

src/main/kotlin/ecommerce/api/ProductController.kt
arekiu: Done! Now the responsibilities of Controller, Service and Repository are clear!

src/main/kotlin/ecommerce/api/ProductService.kt
arekiu: All clear! I understood the functionality and I´m working on it. For now, it has the same attributes than in the Database, but I understand that later I could choose which attributes to show.

src/test/kotlin/ecommerce/api/ProductControllerTest.kt
arekiu: No, it was wrong. Now I´m using SpringBootTest for the controller tests, and Jdbc for services tests (because they are using JdbcProductStore)

src/main/kotlin/ecommerce/api/ProductController.kt
arekiu: Yes! Put is going to replace the existing object, even if some fields are missing (and they will be setted as Null).
Patch replaces the specified attributes in the existing object.

At this moment, the form is showing the current values and it has an attribute "required" to avoid to leave blank fields.
If you suggest to change it to @PatchMapping, I´ll do it!

src/main/kotlin/ecommerce/api/ProductController.kt
arekiu: I read that the most recommended is Constructor Injection because of Immutability and easier testability (I also started to used in the past projects when the reviewer suggested me to implement it).

Field Injection is easy to set but makes the test very difficult.

Setter injection could be better in the case when the dependencies will change or are optional.

src/main/kotlin/ecommerce/api/ProductController.kt
arekiu: I changed my mind and I´m using Patch. I guess, it´s more accurate for this case. But anyway, I realized that is necessary to do some checks in Frontend and Backend to reduce error possibility.

src/main/resources/templates/createProductForm.html
arekiu: Ok, I thought that it was always better to make a get request from the front-end. I didn´t know that it´s a good practice to pass the data through the controller. Thank you!

arekiu:
Good morning/afternoon/evening Dani!

I reflected about some things, made some choices and applied some changes. Let me know what do you think about the current structure!

Have a nice day!

da-nyee (changes requested):
Hi Arekiu! 🙌

I think you might have forgotten to push any commits or changes 👀
Could you check again?

src/test/kotlin/ecommerce/api/ProductServiceTest.kt
```diff
@@ -0,0 +1,92 @@
...
+    fun insert() {
+        val product = Product(name = "Iron body", price = 99.0, imageUrl = "https://alexnsan.comics/imageurl/123")
+        productService.insert(product)
+        val allProducts = productService.findAll()
+        val lastProduct = allProducts.last()
+        assertThat(lastProduct.name).isEqualTo(product.name)
+    }
+
+    @Test
+    fun update() {
```
da-nyee, line 73: image

Could you check this test since it failed?

src/main/kotlin/ecommerce/controller/view/ProductViewController.kt
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.ui.Model
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+
+@Controller
+class ProductViewController(private val productService: ProductService) {
+    @GetMapping("/")
+    fun displayProducts(model: Model): String {
+        val products = productService.findAll()
+        model.addAttribute("products", products.map { ProductDTO.from(it) })
```
da-nyee, line 15: You made it 👍

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    fun update(
+        id: Long,
+        dto: ProductPatchDTO,
+    ): Boolean {
+        val product = productStore.findProductById(id) ?: return false
+
+        dto.name?.let { product.name = it }
+        dto.price?.let { product.price = it }
+        dto.imageUrl?.let { product.imageUrl = it }
```
da-nyee:
```suggestion
        with(dto) {
            apply {
                product.name = name ?: product.name
                product.price = price ?: product.price
                product.imageUrl = imageUrl ?: product.imageUrl
            }
        }
```

You could rewrite the code like this using `with` and `apply` 👀

da-nyee:
```kotlin
product.name = name ?: product.name
product.price = price ?: product.price
product.imageUrl = imageUrl ?: product.imageUrl
```

I also think it would be better to delegate this logic to `Product`.

da-nyee (changes requested):
Hi Arekiu! 🙌
I've gone through all the changes — thanks for pushing them ☺️

You're doing really well so far, but there's just one test failure — could you take another look?
Once it passes, I'll go ahead and merge the PR!

src/test/kotlin/ecommerce/api/ProductServiceTest.kt
arekiu: Ohhh I just forgot to replace it by the DTO :(
Now it´s done!

src/main/kotlin/ecommerce/service/ProductService.kt
arekiu: Moved to Product class. Now I guess that will simplify the validation logic. Thanks!

da-nyee (approved):
Hi Alex! 🙌
Thanks for taking care of the feedback ☺️

Looks like it's good to move on to the next step!
Keep it up 💪 — see you again soon 🤸‍♀️

PR #12, opened 2025-07-17
[spring-ecommerce-product - Step 1-1 to 1-3] Submit Apoorva's mission

Apoorva-S-Natesh:
### **Overview**
This PR introduces CRUD implementation for managing products in an e-commerce application using **Kotlin** and **Spring Boot**, with **Thymeleaf** for the admin interface and **H2** as the database.

### **Product Module**
- [x] Implemented **CRUD HTTP API** via `ProductController` (currently unused in UI).
- [x] Created `ProductViewController` for admin-facing product management using Thymeleaf.
- [x] Added service layer (`ProductService`) to separate business logic from controllers.
- [x] Built repository layer with JDBC and H2 integration.

###  **Testing**
- [x] Added tests for all CRUD operations using RestAssured.
- **Note**: Repository tests may fail when run together due to persistent data in H2. Cleanup strategy or isolated test contexts may be needed.

###  **Admin Interface**
- [x] Admin can view, add, update, and delete products via Thymeleaf templates.
- [x] Buttons added for CRUD actions.
- [ ] Admin login functionality is pending.

### **Database**
- [x] Configured H2 database.
- [x] Created schema and connected repository via service layer.
- [x] Repository uses JDBC with manual SQL queries and row mapping.

### **Design Decisions**
- `ProductController` was initially created for RESTful API access. It is currently not used in the UI since the focus shifted to server-side rendering with Thymeleaf. It will be used in future iterations when JavaScript (e.g., `fetch`) is introduced.
- Business logic is encapsulated in the service layer to maintain clean separation of concerns.
- Repository uses in-memory H2 for simplicity and fast testing.
- We decided to focus more on mapping and the API and to not use JS for this mission

### **Known Issues & Next Steps**
- [ ] Add admin login and authentication.
- [ ] Integrate JavaScript to consume `ProductController` API in future.
- [ ] Add validation and error handling for form inputs.

###Testable Endpoints
/
Redirects to the admin product list page.

/admin/products
Displays a list of all products and a button to add a new product, delete or edit product.

/admin/products/add-product (GET)
Shows the form to add a new product.

/admin/products/add-product (POST)
Handles form submission to create a new product.

/admin/products/edit-product/{id} (GET)
Displays the form to edit an existing product by ID.

/admin/products/edit-product/{id} (POST)
Handles form submission to update an existing product.

/admin/products/delete-product/{id} (DELETE)
Deletes a product by ID and redirects to the product list.

Thank you for reviewing this code.


build.gradle.kts
```diff
@@ -29,6 +29,7 @@ dependencies {
     testImplementation("org.springframework.boot:spring-boot-starter-test")
     testImplementation("org.jetbrains.kotlin:kotlin-test-junit5")
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
+    testImplementation("io.rest-assured:rest-assured:5.3.1")
 }
```
da-nyee, line 33: Do you know the differences between `implementation`, `api`, `runtimeOnly`, and `compileOnly`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.web.bind.annotation.DeleteMapping
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
```
da-nyee, line 19: What is the difference between `@Controller` and `@RestController`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+class ProductController(private val productService: ProductService) {
+    @PostMapping("/products")
+    fun createProduct(
+        @RequestBody newProduct: Product,
+    ): ResponseEntity<Product> {
```
da-nyee: Is it okay to use a model object as a request or a response?
If so, why do you think that’s a good idea? If not, what’s the best practice?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,58 @@
...
+        }
+        return ResponseEntity.ok().body(productService.getProductById(id))
+    }
+
+    @GetMapping("/products")
+    fun getProducts(): ResponseEntity<List<Product>> {
+        return ResponseEntity.ok().body(productService.getAllProducts())
+    }
+
+    @PutMapping("/products/{id}")
```
da-nyee, line 44: Have you heard of `@PatchMapping`?
What is the difference between `@PutMapping` and `@PatchMapping`?

da-nyee: And what about `@PostMapping`?

src/main/kotlin/ecommerce/controller/ProductViewController.kt
```diff
@@ -0,0 +1,82 @@
...
+    @GetMapping("/")
+    fun welcomePage(): String {
+        return "redirect:/admin/products"
+    }
+
+    @GetMapping("/admin/products")
+    fun listAllProducts(model: Model): String {
+        model.addAttribute("products", productService.getAllProducts())
+        model.addAttribute("product", ::Product)
+        return "products"
```
da-nyee, line 24: I think there might be a better way to respond than just returning a plain string.
You could consider returning a status code instead 👀

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,3 @@
+package ecommerce.model
+
+data class Product(var id: Long, var name: String, var price: Double, var img: String, var quantity: Int)
```
da-nyee, line 3: Would it be appropriate to use `data class` for a model object?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,54 @@
...
+                rs.getString("name"),
+                rs.getDouble("price"),
+                rs.getString("img"),
+                rs.getInt("quantity"),
+            )
+        }
+
+    fun count(): Int {
+        val sql = "select count(*) from products"
+        return jdbcTemplate.queryForObject(sql, Int::class.java) ?: 0
```
da-nyee, line 24: Would you feel comfortable using `JdbcTemplate`? Everything working okay?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,25 @@
+package ecommerce.service
+
+import ecommerce.model.Product
+import ecommerce.repository.ProductRepository
+import org.springframework.stereotype.Service
+
+@Service
```
da-nyee, line 7: Have you noticed any difference between `@Component` and `@Service`?

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import io.restassured.RestAssured
+import io.restassured.http.ContentType
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
da-nyee, line 13: Are you familiar with the `classMode` options? How do they differ?

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,25 @@
...
+class ProductService(private val productRepository: ProductRepository) {
```
da-nyee, line 8: You used `Constructor Injection` for DI 👍 
Have you heard of `Field Injection` and `Setter Injection` as well? What are the differences between them?

da-nyee (changes requested):
Hi Apoorva! 🙌
Thanks for waiting for the feedback 🙇‍♀️

You did great work 👍
I've left some questions for you to think about — feel free to check them out.
If anything's unclear, just let me know ☺️

build.gradle.kts
Apoorva-S-Natesh: I wasn't familiar with this section of the `build.gradle.kts`, thank you for bringing it to my attention. I read about it and found the differences.
**Implementation** adds the dependency to both compile and runtime classpaths, not exposed to other modules that depend on this one. Most of our Spring Boot starters (e.g., spring-boot-starter-web, jackson-module-kotlin) are internal to this module and don’t need to be exposed.

 **api** adds the dependency to both compile and runtime classpaths,  but also exposes the dependency to consumers of this module. Since this is an application and not a library we don’t need to expose internal dependencies to other modules.
 
 **runtimeOnly** adds the dependency only at runtime, not at compile time.
H2 database is only needed when the app runs, not during compilation.

**compileOnly** adds the dependency only at compile time, not at runtime.

**testRuntimeOnly** used for test engines or launchers needed only when running tests.

src/main/kotlin/ecommerce/controller/ProductController.kt
Apoorva-S-Natesh: I had this confusion while starting the mission. From my understanding, the difference between `@Controller` and `@RestController` in how they handle HTTP responses.
`@Controller` returns views (like HTML, Thymeleaf pages) and we should use `@ResponseBody` on methods to return raw data (like JSON) and is typically used in Web apps returning webpages.
 `@RestController` Automatically serializes return values (like objects) to JSON and typically used in APIs returning data to clients.
 
 Please let me know if there are other differences and the best way to use them.

src/main/kotlin/ecommerce/controller/ProductController.kt
Apoorva-S-Natesh: While it works as this is a small project it's not ideal for long-term maintainability, I did a little more research and found out that it's not ideal. It exposes the domain model and might leak data unintentionally. This also lacks the flexibility when it comes to accepting fewer fields in request than in return response.
Thank you for bringing this up, I now got to know the best practices of using DTOs(Data Transfer Objects). DTOs help to decouple the API from the domain model, improves security, and provides flexibility for evolving both layers independently.

**What I changed:**
I introduced DTO called: `CreateProductRequest`, `UpdateProductRequest`, and `ProductResponse` that - Accept only the fields needed and return only the fields that can be exposed. extension functions (toModel(), toResponse())- ProductMapper to convert between DTOs and domain models.

I had a confusion between DTO and DAO but please me know if What I think is right:
DTOs transfer data between layers(API) where as DAOs encapsulate database access logic.

src/main/kotlin/ecommerce/controller/ProductController.kt
Apoorva-S-Natesh: Please let me know if this is correct
`@PostMapping`: Used to handle HTTP POST requests, for creating new resources on the server. Each request usually adds something new.
`@PutMapping`: Used for HTTP PUT requests to fully update (or replace) an existing resource. The request body should contain the complete data for the resource; missing fields may be set to null or defaults. Safe to repeat without creating duplicates.
`@PatchMapping`: Used for HTTP PATCH requests to apply partial updates. Only the specified fields in the request are changed; other fields remain as they are. Is it more suitable for complex or large resources when I don't want to resend all data? 

would you suggest using `@PatchMapping` here?

src/main/kotlin/ecommerce/controller/ProductViewController.kt
Apoorva-S-Natesh: Thank you for the feedback — I really appreciate your attention to detail!

In this case, the method is part of a Spring MVC controller that serves HTML views using Thymeleaf. The String returned from the method refers to the name of the view template (e.g., "products"), which Spring resolves and renders as a web page.

Since this isn't a REST endpoint (like in our ProductController), is returning a `ResponseEntity` or an explicit status code necessary? I saw that Spring automatically returns a 200 OK status when the view is successfully rendered.

Is there a way to return a specific status code along with a rendered view?
I tried this and it didnt work:

```
@GetMapping("/admin/products")
fun listAllProducts(): ResponseEntity<ModelAndView> {
    val modelAndView = ModelAndView("products")
    modelAndView.addObject("products", productService.getAllProducts())
    modelAndView.addObject("product", Product(0, "", 0.0, "", 0))

    return ResponseEntity.status(200).body(modelAndView)
}
```

src/main/kotlin/ecommerce/model/Product.kt
Apoorva-S-Natesh: I thought using a data class for a model like Product is appropriate since this class is primarily used to hold data, without complex business logic and also reduces the boilerplate.

I’d love to hear your thoughts, do you see any downsides to using a data class in this context? Or is there a better alternative you’d recommend for this kind of model?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
Apoorva-S-Natesh: Thanks for checking in! So far, everything is working smoothly with JdbcTemplate, and I’m comfortable using it for this project. I appreciate the control it gives over SQL and mapping, especially for a lightweight setup like this with H2.

I’m open to feedback — do you think there’s a better approach for this use case, or something I might be missing?

src/main/kotlin/ecommerce/service/ProductService.kt
Apoorva-S-Natesh: Thank you for asking, I did not find any difference in how Spring handles them.
I think @Service is more expressive. It tells other developers (and Spring tools) that this class belongs to the service layer.
While @Component would work the same technically, @Service adds semantic clarity and aligns with Spring’s recommended layering conventions.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
Apoorva-S-Natesh: Thank you for this question. I have checked 4 `classMode` options within the `@DirtiesContext ` annotation after the lecture from Brie.
**BEFORE_EACH_TEST_METHOD**: The application context is reset before each test method within the class.
**AFTER_EACH_TEST_METHOD**: The application context is reset after each test method within the class.
**AFTER_CLASS**: The application context is reset after the last test method in the class has been executed.
**BEFORE_CLASS**: The application context is reset before the first test method in the class is executed.

src/main/kotlin/ecommerce/service/ProductService.kt
Apoorva-S-Natesh: I saw the differences in a video that using `Setter Injection` like below is used when the dependency needs to be changed after the object creation and when the dependency is optional:
```
@Service
class ProductService() {
    private lateinit var productRepository: ProductRepository
    @Autowired
    fun setProductRepository(productRepository: ProductRepository) {
        this.productRepository = productRepository
    }
    fun getAllProducts(): List<Product> = productRepository.findAllProducts()
```

and `Field Injection` is not ideal for immutability and it could return null in some cases or exception.
```
@Service
class ProductService() {
    @Autowired
    private lateinit var productRepository: ProductRepository
    
    fun getAllProducts(): List<Product> = productRepository.findAllProducts()
```

And Constructor Injection is ideal when the dependency is immutable, and class is always in a valid state (dependencies are provided at creation).

src/main/kotlin/ecommerce/model/Product.kt
Apoorva-S-Natesh: I also have a doubt if overriding the equals and hashcode is required? In that case using data class might not be appropriate when I want to create custom equals/hashcode

As for data class 2 products with same Id but different name and other properties is still different. I think in this case I should consider 2 products with same ID as same

src/main/kotlin/ecommerce/controller/ProductController.kt
da-nyee: @Apoorva-S-Natesh 

> I had a confusion between DTO and DAO but please me know if What I think is right:
DTOs transfer data between layers(API) where as DAOs encapsulate database access logic.

You're right!
One more thing to add — `DAO` is also used to abstract the data access logic, making it easier to manage and test.

src/main/kotlin/ecommerce/controller/ProductController.kt
da-nyee: > would you suggest using `@PatchMapping` here?

Not really.
Since it's not currently used, it might be a bit over-engineered 🤔

src/main/kotlin/ecommerce/controller/ProductViewController.kt
da-nyee: Oh, I think the current return makes sense.
I missed that it's a view controller 😅

Apoorva-S-Natesh:
Hi @da-nyee Thank you very much for your thorough questions! I took my time to look at each one of them, and it really helped me in my learning process. I really appreciate your feedback!

src/main/kotlin/ecommerce/repository/ProductRepository.kt
da-nyee: > I’m open to feedback — do you think there’s a better approach for this use case, or something I might be missing?

Well, I think you're doing great so far.
I don't see anything missing at the moment 👀

src/main/kotlin/ecommerce/model/Product.kt
da-nyee: > I’d love to hear your thoughts, do you see any downsides to using a data class in this context? Or is there a better alternative you’d recommend for this kind of model? 

Well, since we're not building a heavy application or using JPA or anything like that, I think using `data class` to define a model is a good fit.
But as the model grows in complexity, it might be better to consider using just `class` instead.

da-nyee: > I also have a doubt if overriding the equals and hashcode is required? In that case using data class might not be appropriate when I want to create custom equals/hashcode<br/>
As for data class 2 products with same Id but different name and other properties is still different. I think in this case I should consider 2 products with same ID as same

Yeah, I think in a case like that, `class` would be more appropriate than `data class`.
As you know, `data class` compares all properties when generating `equals()` and `hashCode()`.

src/main/kotlin/ecommerce/dto/CreateProductRequest.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.dto
```
da-nyee, line 1: 👍

da-nyee (approved):
Hi Apoorva! 🙌
You've done really well 💯

Thanks for thoughtfully answering the questions — I can see your potential for growth 🌱
I think it's a good time to merge step 1 and move on to step 2.
If anything's unclear, feel free to let me know 🤸‍♀️

PR #13, opened 2025-07-17
[spring-ecommerce-product Step 1] Submit SebasNadu's mission

SebasNadu:
### Summary
This pull request introduces a basic CRUD application using Spring Boot. It includes:
* Two controllers: one for RESTful APIs, another for HTML view rendering.
* A service layer that handles both a fake in-memory database (step 1.1–1.2) and integration with an H2 database(step 1.3).
* Repository layer using JDBC.
* Global exception handling via @ControllerAdvice.
* H2 database configuration with initial schema/data.
* Frontend integration with a simple home page and Bootstrap styling.

### Backend Implementation:

* **Global Exception Handling**: Added `GlobalControllerAdvice` to handle exceptions globally, including a `NotFoundException` for missing model.
* **Controllers**: Implemented `ProductController` for RESTful API endpoints and `ProductViewController` for rendering HTML views.
* **Service and Repository Layers**: Added `ProductService` to handle the fake database and `ProductRepository` for database operations using JDBC templates.

### Database Configuration:

* **Schema and Data Initialization**: Configured H2 database, added schema definition (`schema.sql`), and initialized data (`data.sql`).
* **Application Properties**: Updated `application.properties` for database and development configurations, including enabling the H2 console and live reload(But it's not working for me).

### Frontend Integration:

* **JavaScript CRUD Operations**: Added `main.js` with functions for CRUD operations on products. Added bootstrap CDN for styling.

These changes collectively establish a functional e-commerce platform with a backend API, database integration, and frontend CRUD capabilities.

### Open Questions
1.	JDBC vs Jakarta Persistence (JPA):
Is there a performance benefit in using plain JDBC over Jakarta (JPA)? In which cases is one preferable to the other?
2.	PATCH Endpoint Usage:
We’ve implemented a PATCH method in the backend, but it’s not yet used in the frontend. Should we delay this implementation until it’s actually needed?
3.	Scalability of findAll():
The findAll() method currently returns all products without pagination. As the catalog grows, this might cause memory issues. Should we consider implementing pagination or filtering strategies early on?
4.	Security for deleteAll():
The deleteAll() method is potentially destructive. Should we protect it with authentication or roles to prevent accidental misuse?

Thank you in advance for your review!!


README.md
```diff
@@ -1 +1,25 @@
...
\ No newline at end of file
+# spring-ecommerce-product
+
+## Features
+### Step 1.1
+- [x] Create a Product class
+  - [x] contains id: Long, name: String, price: Double, imageUrl: String
+  - [x] use AtomicLong to create the Id
+- [x] Create ProductController
+  - [x] use @RestController to return always JSON
```
da-nyee, line 9: What's the difference `@Controller` and `@RestController`?

build.gradle.kts
```diff
@@ -19,16 +19,20 @@ repositories {
...
 
+val assureVersion = "5.3.1"
 dependencies {
     implementation("org.springframework.boot:spring-boot-starter-jdbc")
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.h2database:h2:2.2.224") // add
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     runtimeOnly("com.h2database:h2")
```
da-nyee: I'm just curious — why did you add another H2 dependency?
It seems like there's already one defined for the database 👀

da-nyee: I have one more question — do you know the differences between `implementation`, `api`, `runtimeOnly`, and `compileOnly`?

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -0,0 +1,15 @@
+package ecommerce.advice
+
+import ecommerce.exception.NotFoundException
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice
+class GlobalControllerAdvice {
```
da-nyee: You added a controller advice to handle exceptions — nice work 👍

da-nyee: Are there other ways to use `@ControllerAdvice`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.web.bind.annotation.PatchMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
```
da-nyee, line 17: What's the difference between `@Controller` and `@RestController`?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,66 @@
...
+        return ResponseEntity.created(URI.create("$PRODUCT_PATH/${saved.id}")).body(saved)
+    }
+
+    @PutMapping(PRODUCT_PATH_ID)
+    fun updateProductById(
+        @RequestBody product: Product,
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> = ResponseEntity.ok(productRepository.update(id, product))
+
+    @PatchMapping(PRODUCT_PATH_ID)
```
da-nyee, line 41: > PATCH Endpoint Usage:
We’ve implemented a PATCH method in the backend, but it’s not yet used in the frontend. Should we delay this implementation until it’s actually needed?

Yeah, I think this might be somewhat over-engineered, since it hasn't been used yet.
I'd recommend focusing on features that are actually needed for now.

But you know, it's a nice try 👍

da-nyee: Just one question at this point — what's the difference between `@PutMapping` and `@PatchMapping`?

src/main/kotlin/ecommerce/controller/ProductViewController.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+
+@Controller
+@RequestMapping
+class ProductViewController(private val productRepository: ProductRepository) {
+    @GetMapping
+    fun showProducts(model: Model): String {
+        model.addAttribute("products", productRepository.findAll())
+        return "product-list"
```
da-nyee, line 27: Have you heard of status codes?
How about using one instead of just returning a plain string?

src/main/kotlin/ecommerce/exception/NotFoundException.kt
```diff
@@ -0,0 +1,3 @@
+package ecommerce.exception
+
+class NotFoundException(message: String) : RuntimeException(message)
```
da-nyee: You made a custom exception 👍

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    fun partialUpdate(other: Product): Product =
+        this.copy(
+            name = other.name ?: this.name,
+            price = other.price ?: this.price,
+            imageUrl = other.imageUrl ?: this.imageUrl,
+        )
+
+    companion object {
+        fun toEntity(
```
da-nyee: I noticed you're using `Product` as both a DTO and an entity 👀
Have you considered separating them?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,87 @@
...
+import ecommerce.model.Product
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.jdbc.support.GeneratedKeyHolder
+import org.springframework.jdbc.support.KeyHolder
+import org.springframework.stereotype.Repository
+import java.sql.ResultSet
+
+@Repository
+class ProductRepository(private val jdbc: JdbcTemplate) {
```
da-nyee: > JDBC vs Jakarta Persistence (JPA):
Is there a performance benefit in using plain JDBC over Jakarta (JPA)? In which cases is one preferable to the other?

`JDBC` generally offers better performance than `JPA`, but `JPA` has the advantage when it comes to productivity.
Queries are executed directly with `JDBC`, whereas `JPA` operates through object mapping, which introduces some overhead.
`JDBC` also tends to result in duplicated and verbose code, while `JPA` can simplify things with higher-level abstractions.
Overall, it depends on the situation — and in many real-world cases — it's common to use a mix of both.

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,87 @@
...
+        RowMapper<Product> { rs: ResultSet, _ ->
+            Product(
+                rs.getLong("id"),
+                rs.getString("name"),
+                rs.getDouble("price"),
+                rs.getString("imageUrl"),
+            )
+        }
+
+    fun findAll(): List<Product> {
```
da-nyee: > Scalability of findAll():
The findAll() method currently returns all products without pagination. As the catalog grows, this might cause memory issues. Should we consider implementing pagination or filtering strategies early on?

Yes, it could be — but I think it's a bit over-engineered for now.
Still, if you'd like to practice implementing pagination, feel free to go ahead.
I'd be happy to take a look and give you some feedback 👀

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,63 @@
...
+        val updated = existing.partialUpdate(product)
+        products[id] = updated
+        return updated
+    }
+
+    fun delete(id: Long) {
+        if (products.remove(id) == null) throw NotFoundException("Product with Id: $id. Not found.")
+    }
+
+    fun deleteAll() {
```
da-nyee: > Security for deleteAll():
The deleteAll() method is potentially destructive. Should we protect it with authentication or roles to prevent accidental misuse?

Good point — but I think it's also a bit over-engineered for now.
If you have enough time, however, feel free to go ahead and implement the feature 👀

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,63 @@
+package ecommerce.service
+
+import ecommerce.exception.NotFoundException
+import ecommerce.model.Product
+import org.springframework.stereotype.Service
+import java.util.concurrent.atomic.AtomicLong
+
+@Service
```
da-nyee: Is there any difference between `@Component` and `@Service`?

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import io.restassured.RestAssured
+import io.restassured.http.ContentType
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
da-nyee, line 13: Are you familiar with the `classMode` options? Could you explain the differences?

da-nyee (changes requested):
Hi SebasNadu! 🙌
You've been awesome 👍 

I've answered your questions and left a few comments — feel free to check them out.
If anything's unclear, just let me know 🙇‍♀️

README.md
SebasNadu: @Controller is used in the traditional MVC pattern. It returns a view name (for example an HTML page rendered by Thymeleaf).

@RestController is a shortcut for @Controller + @ResponseBody, meaning it returns raw data (like JSON) instead of a view. Use this when you want to build a RestApi.

build.gradle.kts
SebasNadu: Good catch! I’m not exactly sure how the duplicate dependency got in, maybe a teammate added it and I overlooked it, sorry for that!. You’re absolute right, runtime H2 dependency is enough, it was already added with the project, and It's not necessary to use the other one for what we need.

The differences are:
**runtimeOnly()** means the H2 library is available only at runtime, typically used when Spring Boot handles the database configuration for you automatically.
**implementation()** makes the dependency available at compile time as well, which is necessary if you plan to use H2 classes directly in your code, for example, calling org.h2.tools.Server to manually start the server, but we don't need this kind of customization for this project.

For your second question, here’s how I understand the different dependency scopes:
* **implementation** is available at both compile time and runtime. You should use it when your application directly depends on a library in its code. This is the most common case.
* **runtimeOnly** is available only at runtime. Use this when the library is not needed to compile the code, but is required to run the application, for example, a database that Spring Boot auto-configure.
* **compileOnly** is available only at compile time. This is useful when the library is provided at runtime by the environment, such as a servlet container in spring. You use it to compile the code, but you don’t package it in your app.
* **api** What I understand is that is mainly used in big projects with multi modules. It behaves like implementation, but it also exposes the dependency to other modules that depend on your module. This means if Module A depends on Module B, and B declares api("some-lib"), then Module A can also use some-lib without declaring it again. I imagine that could be nice for maintainability and readability because it centralizes the dependencies.

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
SebasNadu: Yes, another way is to use @ExceptionHandler directly inside the controller where the exception occurs, instead of using @ControllerAdvice globally. But the ControllerAdvice maintain the code DRY and centralize all the @Exception handlers🙌.

src/main/kotlin/ecommerce/controller/ProductController.kt
SebasNadu: @controller is used in the traditional MVC pattern. It returns a view name (for example an HTML page rendered by Thymeleaf).

@RestController is a shortcut for @controller + @responsebody, meaning it returns raw data (like JSON) instead of a view. Use this when you want to build a RestApi.

src/main/kotlin/ecommerce/controller/ProductController.kt
SebasNadu: You use a PUT request when you’re updating the entire resource, it’s meant to replace the existing resource completely.
On the other hand, you use PATCH when you want to update only part of the entity, such as a single field or a group of fields.

src/main/kotlin/ecommerce/controller/ProductViewController.kt
SebasNadu: I'm returning a view name here "product-list) since this controller serves a server-rendered html page. If this were a RestApi, then yes, returning a ResponseEntity with an explicit status code would be the way to go.
But now what I'm thinking is what happen if findAll() fails, so maybe I have to throw an specific exception and return an error page. I'll check this 👍.

src/main/kotlin/ecommerce/model/Product.kt
SebasNadu: Yes, I kept things simple for this step since we didn’t need validation or a complex database schema yet. I focused on learning just what was necessary to complete the task.

But you’re right, I’ve also read that Spring commonly separates DTOs and entities. If I understand correctly, the DTO (Data Transfer Object) is meant for sending and receiving data between the client and the server, while the Entity is used for database interaction. The entity may include more fields, validation rules, relationships, and behaviour specific to the db. But I was also confused because to really use @Entity you need to use JPA and not JDBC like the requirements say, right? So also I was insecure if this worth it or not. But I suppose that for the next step with validation, we need to use validations, and it has sense to put inside the Entity and separate this two behaviours.

Thanks again for pointing it out, I’ll definitely try to apply this separation to be better prepared for the next step. 🙂
Let me know if I’m misunderstanding anything!

src/main/kotlin/ecommerce/repository/ProductRepository.kt
SebasNadu: Thank you! You answered my question very well, all clear!

src/main/kotlin/ecommerce/service/ProductService.kt
SebasNadu: I didn’t really know much about @Component before, thanks for asking!

Basically, @Component is a general-purpose annotation for Spring beans that don’t really fit into service or repository roles of the Layered Architecture from Spring. It’s good for things like utility classes.

@Service and @Repository are just more specific kinds of @Component that show what role the class plays in the app, like business logic or data access.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
SebasNadu: Thanks for asking, I didn't know so much about the other modes before 👍.

AFTER_EACH_TEST_METHOD: Resets the context after every test method, It ensures a complete isolation in each test, but it has an extra cost in the performance.
BEFORE_EACH_TEST_METHOD very similar to the previous one. Reset the context before every test method.
AFTER_CLASS Reset the context after all methods in a class. It's faster and isolate the tests per class.
BEFORE_CLASS the same as the previous one bet it reset the context before not after the class initiate.
AFTER_TEST_METHOD Reset the context after a specific test method.

Let me known if I misunderstood anything please!

SebasNadu:
# Recent Changes Summary

## Backend Improvements
- Implemented `ProductDTO`, `Product` entity, and `ProductPatchDTO`.
- Added extension functions for mapping between DTO and entity both ways.
- Created a `ProductService` interface with two implementations:  
  - `ProductServiceJDBC` handles repository interaction, business logic, and DTO-entity transformations.  
  - `ProductServiceCollection` works with an in-memory collection as a mock DB.  
- Moved product persistence logic to `ProductRepository` focusing solely on the entity, keeping service and repository responsibilities clear.  
- Used `@Primary` to specify the default service implementation to inject.

## Error Handling and Stability
- Added an `ApiErrorController` with `@RestControllerAdvice` to extend exception handling for API endpoints.  
- Added a `ViewErrorController` with `@ControllerAdvice` for custom error pages and consistent feedback in the UI.  
- Expanded custom exceptions to cover service, repository, and JDBC-specific errors, improving error granularity and handling.

## Features and Enhancements
- Added pagination support for product listings, with backend logic and UI updates for better user experience and scalability.  
- Refactored `ProductViewController` and associated views to handle paginated results.  
- Implemented modal dialogs for adding and editing products, streamlining interactions without page reloads.  
- Improved product deletion flow, supporting both bulk and individual deletes with clear user notifications.  
- Integrated Font Awesome icons for edit and delete buttons to enhance UI clarity and usability.

CleanShot 2025-07-21 at 00 07 36@2x

src/main/kotlin/ecommerce/controller/ProductController.kt
da-nyee: @SebasNadu 

Ohhh I asked it twice 😅
Thanks for the thoughtful answer 🙇‍♀️

src/main/kotlin/ecommerce/model/ProductDTO.kt
```diff
@@ -0,0 +1,24 @@
+package ecommerce.model
+
+data class ProductDTO(
```
da-nyee, line 3: Good separation between DTO and Entity 👍

src/main/kotlin/ecommerce/services/ProductService.kt
```diff
@@ -0,0 +1,28 @@
+package ecommerce.services
+
+import ecommerce.model.ProductDTO
+import ecommerce.model.ProductPatchDTO
+
+interface ProductService {
```
da-nyee, line 6: Nice try 👍
You built two types of DAO(Data Access Object) — one using a collection, and the other using JDBC.

src/main/kotlin/ecommerce/services/ProductServiceCollection.kt
```diff
@@ -0,0 +1,75 @@
+package ecommerce.services
+
+import ecommerce.exception.NotFoundException
+import ecommerce.model.ProductDTO
+import ecommerce.model.ProductPatchDTO
+import org.springframework.stereotype.Service
+import java.util.concurrent.atomic.AtomicLong
+
+@Service
+class ProductServiceCollection : ProductService {
```
da-nyee, line 10: By the way, how about annotating this class with `@Component` or `@Repository`?
Since it mainly deals with data rather than business logic, one of those might be more appropriate.

src/main/kotlin/ecommerce/services/ProductServiceJDBC.kt
```diff
@@ -0,0 +1,66 @@
...
+import ecommerce.mappers.toEntity
+import ecommerce.model.ProductDTO
+import ecommerce.model.ProductPatchDTO
+import ecommerce.repositories.ProductRepository
+import org.springframework.context.annotation.Primary
+import org.springframework.stereotype.Service
+
+@Service
+@Primary
+class ProductServiceJDBC(private val productRepository: ProductRepository) : ProductService {
```
da-nyee, line 15: This one too 👀

src/main/kotlin/ecommerce/services/ProductServiceJDBC.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun findAll(): List<ProductDTO> = productRepository.findAll().map { it.toDTO() }
+
+    override fun findAllPaginated(page: Int, size: Int): Pair<List<ProductDTO>, Int>  {
```
da-nyee, line 18: You implemented pagination 👍

da-nyee: Here's one more thing to consider — have you heard of offset-less pagination, also known as cursor-based pagination?
Why do you think this approach might improve performance?

src/main/kotlin/ecommerce/util/Logging.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.util
+
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+inline fun <reified T : Any> logger(): Logger = LoggerFactory.getLogger(T::class.java)
```
da-nyee, line 6: You used `inline` 👍

da-nyee: What specific features does `inline` have?

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,147 @@
...
+class ProductControllerTest {
```
da-nyee, line 14: How about adding tests for the `ProductService` implementations?

src/main/kotlin/ecommerce/advice/ApiErrorControllerAdvice.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.dao.DuplicateKeyException
+import org.springframework.dao.EmptyResultDataAccessException
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.bind.annotation.RestControllerAdvice
+
+@RestControllerAdvice(annotations = [RestController::class])
+class ApiErrorControllerAdvice {
```
da-nyee, line 16: 👍

src/main/kotlin/ecommerce/advice/ViewErrorControllerAdvice.kt
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.dao.DuplicateKeyException
+import org.springframework.dao.EmptyResultDataAccessException
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.stereotype.Controller
+import org.springframework.ui.Model
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice(annotations = [Controller::class])
+class ViewErrorControllerAdvice {
```
da-nyee, line 16: 👍

da-nyee (approved):
Hi SebasNadu! 🙌
You did a really great job 💯

Thanks for carefully answering the questions ☺️
You've improved the code nicely — looks like you're already comfortable with the Spring Framework.

I think step 1 is good to go, so I'll go ahead and merge it.
Keep going with step 2, and feel free to reach out if you need any help 💪

PR #14, opened 2025-07-17
[Spring-ecommerce-product - Step1] Submit codebrew42's mission

jieun-pak:
Hello,  안녕하세요! 😊

This pull request introduces a full implementation of a Spring-based e-commerce product management system, including backend services, H2 database integration, and frontend JavaScript enhancements for basic CRUD functionality.

As this was my first time working on a web development project, I found it challenging but also very enjoyable because I’m genuinely interested in the topic. Thank you in advance for your opinion and feedback!

---

### Focus

Our main goal was to implement the core requirements as clearly as possible. We intentionally didn’t spend much time polishing the JavaScript side, since we treated it more like a frontend utility than the main focus.

---

### Challenges

Testing was one of the difficult parts, especially because understanding the boundary between frontend and backend and how to verify each piece.

---

### Questions

1. **Testing Web Apps**
   What makes testing web development different than projects like Lotto or Blackjack apps?

2. **SQL Insertion with Absolute Paths**
   I initially tried to use an absolute path for SQL data insertion via H2 like below for learning purpose. In what scenarios is this approach appropriate?

   ```
   spring.datasource.url=jdbc:h2:file:/Users/jieun.park.3/Downloads/coding/spring-ecommerce-product-sebsrepo-learning1/data
   ```

3. **Repository vs JPA**
   What should we consider when choosing between using `JdbcTemplate` (like our `ProductRepository`) vs Spring Data JPA with `@Entity`?


build.gradle.kts
```diff
@@ -19,16 +19,20 @@ repositories {
     mavenCentral()
 }
 
+val assureVersion = "5.3.1"
 dependencies {
     implementation("org.springframework.boot:spring-boot-starter-jdbc")
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.h2database:h2:2.2.224") // add
```
Wordbe: Why did you add this dependency? Is there a specific purpose or problem it’s solving? 🙂

build.gradle.kts
```diff
@@ -19,16 +19,20 @@ repositories {
     mavenCentral()
 }
 
+val assureVersion = "5.3.1"
```
Wordbe, line 22: 👍

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
```diff
@@ -0,0 +1,15 @@
+package ecommerce.advice
+
+import ecommerce.exception.NotFoundException
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice
```
Wordbe, line 11: I think you would prefer a JSON response—am I right? 🙂

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
+@RequestMapping
```
Wordbe: This doesn’t seem necessary!

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,66 @@
...
+    @DeleteMapping(PRODUCT_PATH_ID)
+    fun deleteProductById(
+        @PathVariable id: Long,
+    ): ResponseEntity<Unit> {
+        productRepository.delete(id)
+        return ResponseEntity.noContent().build()
+    }
+
+    @DeleteMapping(PRODUCT_PATH)
+    fun deleteAllProducts(): ResponseEntity<String> {
```
Wordbe: Is returning a string really necessary here? 🙂

src/main/resources/schema.sql
```diff
@@ -0,0 +1,8 @@
+DROP TABLE IF EXISTS TBL_PRODUCTS;
+
+CREATE TABLE TBL_PRODUCTS (
```
Wordbe: There’s no absolute rule, but table names are usually lowercase. Also, how about removing the `TBL_` prefix from table names? 🙂

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,87 @@
...
+        val sql = "insert into TBL_PRODUCTS (name, price, imageUrl) values (?, ?, ?)"
+
+        val keyHolder: KeyHolder = GeneratedKeyHolder()
+        jdbc.update({
+            it.prepareStatement(sql, arrayOf("id")).apply {
+                setString(1, product.name)
+                setDouble(2, product.price!!)
+                setString(3, product.imageUrl)
+            }
+        }, keyHolder)
```
Wordbe: Since you registered Auto Increment for the ID in `schema.sql` for the H2 Database, the database will automatically generate incrementing IDs.
Therefore, there’s no need to create or insert the ID manually here!

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    fun delete(id: Long) {
+        val sql = "DELETE FROM TBL_PRODUCTS WHERE ID = ?"
+        jdbc.update(sql, id)
+    }
+
+    fun deleteAll() {
+        val sql = "DELETE FROM TBL_PRODUCTS"
+        jdbc.update(sql)
+    }
```
Wordbe: This is a very risky query and probably wouldn’t be used in real scenarios, but I’ll consider it as practice!

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,63 @@
+package ecommerce.service
+
+import ecommerce.exception.NotFoundException
+import ecommerce.model.Product
+import org.springframework.stereotype.Service
+import java.util.concurrent.atomic.AtomicLong
+
+@Service
+class ProductService {
```
Wordbe: You’ve created the Service but aren’t using it yet.
Have the Service depend on the Repository, and let the Controller use the Service instead of directly accessing the Repository! 🙂

src/main/resources/application.properties
```diff
@@ -1 +1,19 @@
+spring.datasource.url=jdbc:h2:mem:testdb
+spring.datasource.username=sa
+spring.datasource.password=
+
+spring.sql.init.mode=always
+spring.jpa.show-sql=true
+spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.H2Dialect
+spring.jpa.hibernate.ddl-auto=create
```
Wordbe: Since you’re not using JPA, please remove the JPA-related configurations!

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,142 @@
...
+import io.restassured.RestAssured
+import io.restassured.http.ContentType
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
Wordbe, line 13: Could you explain why this is necessary? What’s the reason behind it? 🤔

Wordbe (changes requested):
Hello Jin!

Nice to meet you!
Welcome to the Spring Boot project mission.
Looking at your code, it’s clear you have a good understanding of controllers and domain objects, and the structure is well done.

I’ve left some review comments with points to consider—please check them and apply any changes you can.
Thank you! 😊

---

Q1. Testing Web Apps
What makes testing web development different than projects like Lotto or Blackjack apps?

Web app testing requires comprehensive tests that consider user experience and the overall system flow. Since external network connections are heavily involved, designing your system to control or mock these parts will help make testing easier and more reliable.
On the other hand, apps like Lotto or Blackjack focus more on verifying business logic, which can be similar to domain object testing even within web apps. 🙂

Q2. SQL Insertion with Absolute Paths
I initially tried to use an absolute path for SQL data insertion via H2 like below for learning purpose. In what scenarios is this approach appropriate?

When learning or working on personal projects, using an absolute path to fix the database location on your PC is fine.
However, in a real service environment, it’s better to manage test data flexibly—using relative paths, configuration files, or database settings—to adapt to different deployment setups. 

Q3. Repository vs JPA
What should we consider when choosing between using JdbcTemplate (like our ProductRepository) vs Spring Data JPA with @Entity?

JPA provides a high level of abstraction, allowing developers to use it conveniently and boost productivity. However, a solid conceptual understanding is essential, and for complex queries like batch inserts, JDBC may still be necessary.
On the other hand, JDBC operates at a lower level, requiring direct SQL handling. This gives developers more flexibility and freedom to write custom queries as needed.
Consider the pros and cons of each when making your choice! 😊

Thank you!

build.gradle.kts
jieun-pak: I realized having this line is enough, and Spring Boot manages the version.
```
```
I deleted that line, since I don't  have any specific need to specify the version

And I could learn how to check the dependency tree for the runtimeClasspath configuration with a filter, thanks to your comment.
```
./gradlew dependencies --configuration runtimeClasspath   
 ./gradlew dependencies --configuration runtimeClasspath | grep h2
```

build.gradle.kts
jieun-pak: I'll keep in mind to use it, since Gradle's Kotlin DSL (build.gradle.kts) encourages using `val` declarations for better centralized version management. Thanks!

src/main/resources/application.properties
jieun-pak: They were added while learning about it, I removed it. Thanks!

src/main/kotlin/ecommerce/controller/ProductController.kt
jieun-pak: I removed unnecessary "@RequestMapping", thanks!

src/main/resources/schema.sql
jieun-pak: I did incorporate your feedback, thanks!

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
jieun-pak: Yes, I implemented JSON response. Thank you!

src/main/kotlin/ecommerce/controller/ProductController.kt
jieun-pak: I standardized return types, and replaced it with ResponseEntity<Unit>! :)

src/main/kotlin/ecommerce/repository/ProductRepository.kt
jieun-pak: I'll keep in mind, and try to make it less dangerous next time! thanks!

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
jieun-pak: @SpringBootTest:
  Spring Boot Test starts the actual Spring Boot application with a real HTTP server (on port 8080). RestAssured can then make real HTTP calls like GET localhost:8080/api/products. In short, this annotation enables integration testing of everything together : controller, database, and all components.

  @DirtiesContext:
  My test methods create/modify/delete products independently. Without resetting the application context, previous test data would interfere with subsequent tests, causing failures. For example, if one test creates a product, the next test expecting exactly 5 products would find 6 instead. That's why I need to use @DirtiesContext to ensure each test starts with a clean state.

src/main/kotlin/ecommerce/service/ProductService.kt
jieun-pak: After replacing this, I didn't remove this class, since I didn't fully understand the process of replacing in-memory database with the real database-logic. I will think about the naming. Thanks!

src/main/kotlin/ecommerce/repository/ProductRepository.kt
jieun-pak: I misunderstood and thought I needed to handle the ID manually, but I just learned that the database generates IDs automatically. I've updated the code based on that, and it's much cleaner now. Thank you!

```
 schema.sql:
  CREATE TABLE products (
      id BIGINT AUTO_INCREMENT PRIMARY KEY,  ← here
      ...
  )
```

jieun-pak:
Thank you for your comment and for the clear, thoughtful answer to my earlier question.
I took your feedback seriously, especially the part about having a delete-all-products in our code!

In addition to that, I made several improvements:
- Standardized return types to better follow REST conventions
- Removed manual ID handling, since it's auto-generated
- Improved overall readability
- Renamed the table to use lowercase for consistency

I really appreciate your feedback! 🙏

build.gradle.kts
Wordbe: Yeah Great 🙂

build.gradle.kts
Wordbe: You even looked up how to check runtime dependencies with `runtimeClasspath`. Great job! Thank you 😊

src/main/kotlin/ecommerce/advice/GlobalControllerAdvice.kt
Wordbe: You can use `@RestControllerAdvice` and in this case you can omit the `ResponseEntity` and return just the body.

src/main/kotlin/ecommerce/repository/ProductRepository.kt
Wordbe: Sounds good !

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
Wordbe: You're understanding it well. Thanks for looking it up 😊

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    @PutMapping("/{id}")
+    fun updateProductById(
+        @RequestBody product: Product,
+        @PathVariable id: Long,
+    ): Product = productRepository.update(id, product)
+
+    @PatchMapping("/{id}")
+    fun patchProductById(
+        @RequestBody patchRequest: ProductPatchRequest,
```
Wordbe, line 45: In the next step, try using DTOs in other parts(POST, PUT) as well instead of only here 🙂

Wordbe (approved):
Hello Jin,

Thank you for carefully reading through the review, understanding it well, and thoughtfully applying the feedback 😊
I’ll go ahead and approve and merge it from here. Great job!

As for the additional comments I left, please consider reflecting them in the next step if possible.

You did an excellent job on this mission, so I’m confident you’ll do just as well in the next one.
See you in the next step. Thank you!

src/main/kotlin/ecommerce/controller/ProductController.kt
jieun-pak: I'll keep in mind, thank you!

PR #15, opened 2025-07-17
[Spring-Ecommerce-Product Step 1] Submit miooo0o's mission

miooo0o:
This PR implements a basic application with Spring Boot + Kotlin.

Given the project timeline and priorities, we focused on **minimum viable requirements** and prioritized core functionality implementation.

For the frontend components (HTML/JavaScript), we utilized AI tools to generate initial code for development speed, then modified as needed.

## Key Implementation
- operations (Create, Read, Update, Delete)
- H2 in-memory database integration
- Basic web UI
- test code

---
I am satisfied with having at least functional code up and running, but the current implementation basic level and the following areas need additional work:

- Enhanced input validation
- Error handling and unified response format
- and more...

### Please provide feedback on this implementation.

- I'm aware that this implementation is extremely minimal. I'd like to understand better what would constitute "true minimum requirements".
- Additionally, given my limited understanding of Spring, I'm curious about what should be prioritized when developing the code further.
- How to test code? if you have any tips, let me know!


src/main/kotlin/ecommerce/product/Product.kt
```diff
@@ -0,0 +1,18 @@
+package ecommerce.product
+
+class Product(var id: Long? = null, var name: String, var price: Double, var imageUrl: String?) {
```
Wordbe: If the constructor gets too long, it’s better to split it into multiple lines for easier readability! 🙂
https://kotlinlang.org/docs/coding-conventions.html#class-headers

src/main/kotlin/ecommerce/product/Product.kt
```diff
@@ -0,0 +1,18 @@
...
+    }
+
+    companion object {
+        fun toEntity(
+            product: Product,
+            id: Long,
+        ): Product {
+            return Product(id, product.name, product.price, product.imageUrl)
+        }
+    }
```
Wordbe: In my opinion, the method name `toEntity` feels like converting an object into an Entity, so another method name might be more suitable.
Alternatively, using named arguments in the constructor to create the object could also work well!

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val index = AtomicLong(1)
+
+    @PostMapping("/api/products")
+    fun create(
+        @RequestBody product: Product,
+    ): ResponseEntity<Product> {
+        val id = index.getAndIncrement()
+        val newProduct = Product.toEntity(product, id)
+
+        repository.insert(id, newProduct)
```
Wordbe: Since you registered Auto Increment for the ID in `schema.sql` for the H2 Database, the database will automatically generate incrementing IDs.
Therefore, there’s no need to create or insert the ID manually here! 🙂

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.web.bind.annotation.RequestBody
+import java.net.URI
+import java.util.concurrent.atomic.AtomicLong
+
+@Controller
+class ProductController(private val repository: ProductRepository) {
+    private val index = AtomicLong(1)
+
+    @PostMapping("/api/products")
+    fun create(
```
Wordbe: When using the `@Controller` annotation to create APIs that return JSON, you need to add `@ResponseBody` on the methods. 

For reference: try studying `@RestController` as well—it simplifies creating JSON-returning APIs by combining `@Controller` and `@ResponseBody` 🙂

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+        product: Product,
+    ): Int? {
+        val sql =
+            """
+            UPDATE products 
+            SET name = ?, price = ?, image_url = ?
+            WHERE id = ?
+            """.trimIndent()
+
+        return jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, id).takeIf { it == 1 }
```
Wordbe: How about changing the return type to clearly indicate whether the update was successful or not?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    fun deleteById(id: Long): Int? {
+        if (isEmptyOrNull()) return null
+        val sql = "DELETE FROM products WHERE id = ?"
+        return jdbcTemplate.update(sql, id).takeIf { it == 1 }
```
Wordbe: Same goes here as well! 🔄

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+            UPDATE products 
+            SET name = ?, price = ?, image_url = ?
+            WHERE id = ?
+            """.trimIndent()
+
+        return jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, id).takeIf { it == 1 }
+    }
+
+    fun deleteById(id: Long): Int? {
+        if (isEmptyOrNull()) return null
```
Wordbe: I don’t think this part is necessary. If there’s no data matching the ID, the delete query won’t remove anything anyway.
If you want to be safe, how about doing a `SELECT ... WHERE ID = id` instead?
Using `count` can cause unnecessary load and degrade performance! 🙂

src/main/resources/application.properties
```diff
@@ -1 +1,5 @@
 spring.application.name=spring-ecommerce
+
+spring.h2.console.enabled=true
+spring.h2.console.path=/h2-console
+spring.datasource.url=jdbc:h2:mem:database
```
Wordbe, line 5: You’ve created an `application.yml` file—try moving just this file under the `resources` folder and use it.
For Spring Boot configuration, you only need either `application.properties` or `application.yml`, not both.

src/main/resources/schema.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE products
+(
+    id        BIGINT PRIMARY KEY AUTO_INCREMENT,
+    name      VARCHAR(255),
+    price DOUBLE,
+    image_url TEXT
```
Wordbe: `TEXT` can be quite large. Check the actual length you need and consider changing it to a `VARCHAR` with an appropriate length.

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+
+import io.restassured.RestAssured
+import io.restassured.http.ContentType
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
Wordbe: Take a moment to consider why this part is necessary!

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+
+        val response =
+            RestAssured
+                .given().log().all()
+                .contentType(ContentType.JSON)
+                .`when`().get("/api/products")
+                .then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.OK.value())
+        assertThat(response.jsonPath().getList("", Product::class.java)).hasSize(1)
```
Wordbe: Create precise tests by comparing each response value one by one!

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+
+        val AMERICANO =
+            Product(
+                name = "Iced Americano T",
+                price = 4.50,
+                imageUrl =
+                    "https://st.kakaocdn.net/product/gift/product/20231010111814_9a667f9eccc943648797925498bdd8a3.jpg",
+            )
+    }
+}
```
Wordbe: That’s a good approach! The `Product` objects created here for testing are called test fixtures.
To keep responsibilities clear, how about moving these fixtures to a separate class? 🙂

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
```diff
@@ -0,0 +1,164 @@
...
+
+        val response =
+            RestAssured
+                .given().log().all()
+                .contentType(ContentType.JSON)
+                .`when`().get("/api/products")
+                .then().log().all().extract()
+
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.OK.value())
+        assertThat(response.jsonPath().getList("", Product::class.java)).hasSize(2)
```
Wordbe: Please write more detailed tests for other parts as well, not just here! 🙂

Wordbe (changes requested):
Hello Mina!

Welcome to your first Spring Boot project mission. Nice to meet you!
You might have some concerns about Spring, but if you take it step by step, you’ll quickly become familiar and comfortable with it.

I’ve left some review comments—please check them out and apply any changes you can.

And It looks like the Edit and Delete functions aren’t working on the HTML page. Please double-check them! ⚠️

Thank you! 😊

---

Q1. I'm aware that this implementation is extremely minimal. I'd like to understand better what would constitute "true minimum requirements".

I think the minimum requirement for this mission seems to be whether you can perform CRUD operations on products! 

Q2. Additionally, given my limited understanding of Spring, I'm curious about what should be prioritized when developing the code further.

Try to dive deeply into the fundamental principles behind all the code you’re writing or need to write, asking yourself questions about each part whenever possible!
This approach is probably the most effective way to learn. 

Q3. How to test code? if you have any tips, let me know!

If testing feels difficult, try thinking of testing *every* method you’ve created thoroughly.
Also, consider how to design your code structure to make testing easier — that mindset will naturally lead to better code architecture as well. 🙂

src/main/resources/application.properties
miooo0o: Oh, I was wondering why those two even seemed similar, but I wasn't need! Then which one should I save? It seems like it's the `application.properties`, right? I'll look it up more

src/main/kotlin/ecommerce/repository/ProductRepository.kt
miooo0o: The initial goal was to mimic the behavior of a method that used a Map<>, so I agree the return value feels a bit off. A Boolean would probably make more sense.
....
However, during refactoring, I realized I was already returning a response status along with a body, so it made more sense to return the full `ResponseEntity` and convert the model to a response accordingly.

Let me know if you think it still feels off, or if I should consider another structure.

```kotlin
  fun update(
        id: Long,
        request: ProductRequest,
    ): Product? {
        if (!existsById(id)) return null
        val sql = ConstantsSQL.UPDATE_BY_ID.trimIndent()
        val entity = request.toEntity(id)
        val result = jdbcTemplate.update(sql, entity.name, entity.price, entity.imageUrl, entity.id) == 1
        return if (result) entity else null
    }
```

src/main/kotlin/ecommerce/product/ProductController.kt
miooo0o: I didn’t realize how the conversion to JSON worked, but it turns out it was related to the controller, right? After looking into the annotation you mentioned, it makes a bit more sense now. I’ll apply the change!

---

`@RestController`: default JSON returning
`@Controller`:  default HTML returning
`@ResponseBody`: to convert JSON format

src/main/kotlin/ecommerce/repository/ProductRepository.kt
miooo0o:
```kotlin
    fun deleteById(id: Long): Boolean {
        if (!existsById(id)) return false
        val sql = "DELETE FROM products WHERE id = ?"

        return jdbcTemplate.update(sql, id) == 1
    }

    private fun existsById(id: Long): Boolean {
        val sql = "SELECT count(*) FROM products WHERE id = ?"
        val found = jdbcTemplate.queryForObject(sql, Int::class.java, id)
        return found != null && found > 0
    }
```
What do you think about this approach?
Do you think it's really necessary, or is it overkill in most cases?

src/main/kotlin/ecommerce/repository/ProductRepository.kt
miooo0o: I applied it as you said!

src/main/resources/schema.sql
miooo0o: I applied the validation as you suggested!
After a bit of research, I set the limit to 255 characters.

But I'm curious, what would you consider an ideal length for this kind of field(for this mission), and why?

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
miooo0o: `@DirtiesContext(classMode = BEFORE_EACH_TEST_METHOD)` **resets** the `ApplicationContext` (the container that manages Spring Boot beans) before each test runs, so that any shared state like beans or DB changes won’t affect the next test.

This makes sure each test runs in a fully isolated environment, helping to keep the tests independent and reliable.

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
miooo0o: how about this approach?


```kotlin
  val product =
            response.body().`as`(ProductResponse::class.java)
        assertThat(response.statusCode()).isEqualTo(HttpStatus.CREATED.value())
        assertProductEquals(product, AMERICANO, 1)

...

        fun assertProductEquals(
            actual: ProductResponse,
            expected: ProductRequest,
            expectedId: Long,
        ) {
            assertThat(actual.id).isEqualTo(expectedId)
            assertThat(actual.name).isEqualTo(expected.name)
            assertThat(actual.price).isEqualTo(expected.price)
            assertThat(actual.imageUrl).isEqualTo(expected.imageUrl)
        }
```

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
miooo0o: good idea! Thank you.
I gather all to `object TextFixture`But still, I didn't really organize well, if you have any suggestion please let me know

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
miooo0o: I believe this is related to [the previous comment](https://github.com/herotechcourse/spring-ecommerce-product/pull/...).
I added more tests and refactored the code based on your suggestions!

miooo0o:
Hi, Suno. 

The following tasks are based on the feedback received in [PR #15](https://github.com/herotechcourse/spring-ecommerce-product/pull/15). These improvements aim to enhance functionality, readability, and test reliability.

#### in main
- [x] remove application.yml
- [x] Fix Edit/Delete button issue
- [x] Improve constructor readability
- [x] Rename or refactor `Product.toEntity`
- [x] Check/Read about Entity and DTO
- [x] Remove manual ID generation
- [x] Adjust controller for JSON responses
- [x] Clarify update return type
- [x] Optimize database column types
#### in test:
- [x] Review `@DirtiesContext` usage
- [x] Verify each response field
- [x] Separate test fixtures
- [x] Add tests for uncovered features
    
---

### Additional change
- [x] Change price to BigDecimal
- [x] Create an object to collect all SQL lines, `sql/ConstantsSQL`
- [x] Separate `Product` class to `Product`, `ProductRequest`, and `ProductResponse`

###

While refactoring to let the database auto-generate IDs via the schema, I realized that the request object should not include an ID at all. This exposed an ambiguous state in my code, where some objects carried an ID and others didn’t.

I started thinking about how to properly separate these concerns in the Product model, and remembered your earlier comment about the use of "entity." I discussed the concept of DTO vs Entity with other crew members, and concluded that separating them would be necessary sooner or later.

So I introduced a basic separation using extension functions for now, keeping it minimal and easy to change later.

As a result, I also had to update both the `ProductAPI` and the `Repository`. During that process, I revisited your comment about returning Boolean from some repository methods, and applied that suggestion as well.

This led me to reconsider the purpose of each method in the context of `HTTP/1.1`. I spent some time reviewing the specification and realized that some of my methods should return a response body. So I modified them to return entities where appropriate.

```kotlin
fun update(id: Long, request: ProductRequest): Product? 
fun create(request: ProductRequest): Product
```


Thank you for your time :) !

src/main/resources/application.properties
Wordbe: Using `application.yaml`, or YAML files in general, can improve readability 😊 Of course, this is a subjective matter. You can try both and go with whichever works better for you. But when you join a team, it’s usually best to follow the team’s conventions.

src/main/kotlin/ecommerce/product/ProductController.kt
Wordbe: Yes, you understood it well. Thank you 😊

src/main/kotlin/ecommerce/repository/ProductRepository.kt
Wordbe: I think throwing an exception instead of returning null when an update fails is more natural 😊 Failing fast can help prevent unpredictable issues from occurring.

src/main/resources/schema.sql
Wordbe: Most browsers like Chrome and Edge recommend a length of around 2000 characters 😊 So I think that should be fine!

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
Wordbe: Sounds good, thank you 😊

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
Wordbe: I think it’s fine to just list them out without extracting them into separate methods 😊

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
Wordbe: Did you mean `TestFixture` instead of `Text`? 😊 It looks like you did a good job creating it as an object!

src/main/kotlin/ecommerce/product/Product.kt
```diff
@@ -0,0 +1,19 @@
...
+import java.math.BigDecimal
+
+data class Product(
+    val id: Long,
+    val name: String,
+    val price: BigDecimal,
+    val imageUrl: String,
+)
+
+fun Product.toResponse(): ProductResponse {
```
Wordbe, line 12: I think extension functions are best used only when truly necessary 😊 In this case, it seems like it could be implemented as an internal method instead.

From another perspective, the hierarchical structure used in this project is easier to manage when dependencies flow in one direction. For example, a `Controller` can know about a `Model`, but the `Model` shouldn't know about the `Controller`. Currently, `Product` is aware of `ProductResponse`. It would be better to move this method out so that the domain model doesn't depend on a higher layer.

src/main/kotlin/ecommerce/product/Product.kt
```diff
@@ -0,0 +1,19 @@
+package ecommerce.product
+
+import java.math.BigDecimal
+
+data class Product(
```
Wordbe, line 5: What’s the reason for using a `data class` here? 🙂

src/main/kotlin/ecommerce/product/ProductRequest.kt
```diff
@@ -0,0 +1,16 @@
+package ecommerce.product
+
+data class ProductRequest(
```
Wordbe, line 3: Same here. What's the reason for using data class?

src/main/kotlin/ecommerce/product/ProductRequest.kt
```diff
@@ -0,0 +1,16 @@
...
+    val name: String,
+    val price: String,
+    val imageUrl: String,
+)
+
+fun ProductRequest.toEntity(id: Long): Product {
```
Wordbe, line 9: I personally think it would be better to make it an internal method—was there a specific reason you chose to use an extension function instead? 😊

src/main/kotlin/ecommerce/product/ProductResponse.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.product
+
+data class ProductResponse(
```
Wordbe, line 3: Same!

src/main/kotlin/ecommerce/repository/ProductStore.kt
```diff
@@ -0,0 +1,84 @@
...
+        val sql = ConstantsSQL.INSERT.trimIndent()
+        val keyHolder = GeneratedKeyHolder()
+
+        jdbcTemplate.update({ connection ->
+            createPreparedStatement(connection, sql, request)
+        }, keyHolder)
+
+        return keyHolder.key?.toLong()
+            ?: throw IllegalStateException("Failed to retrieve generated ID")
+    }
```
Wordbe, line 71: I thought there was no need to create an id— is there a reason this code is still here?

src/test/kotlin/ecommerce/TextFixture.kt
```diff
@@ -0,0 +1,85 @@
...
+
+    val AMERICANO =
+        ProductRequest(
+            name = "Iced Americano T",
+            price = "4.50",
+            imageUrl =
+                "https://st.kakaocdn.net/product/gift/product/20231010111814_9a667f9eccc943648797925498bdd8a3.jpg",
+        )
+
+    object Dummy {
```
Wordbe, line 40: I think the data above is also "Dummy." It might be better to just remove the Dummy 😊

src/test/kotlin/ecommerce/TextFixture.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        val SUPER_LONG_URL =
+            DummyRequest(
+                name = "Flat white L",
+                price = "6.50",
+                imageUrl = superLongUrl(SCHEMA_SQL_URL_LIMIT),
+            )
+    }
+
+    object AssertTamplate {
```
Wordbe, line 73: Is it appropriate for the Fixture object to also serve as an assert helper? 🙂

Wordbe (approved):
Hello Mina!
I saw a lot of thoughtful work and code in this PR.
You also worked hard researching and studying to create better code. Thank you for your effort.

The structure is well designed, and since most of the mission’s requirements are met, I will approve and merge this PR.
However, I left several comments with feedback, so please take some time to reconsider those points. For the questions I asked, please reply in the comments, and if possible, include any code changes in the next step submission.

Great job! Thank you 😊

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
miooo0o: I misspelled that! I’ll rename it in step 2. Thanks a lot!

src/test/kotlin/ecommerce/product/ProductControllerTest.kt
miooo0o: I applied it as you reviewed!
Wrapping it doesn't seem to make it any better.

src/main/resources/schema.sql
miooo0o: Thank you. I increased it to 2,000 characters!

src/main/kotlin/ecommerce/repository/ProductRepository.kt
miooo0o: I realized during refactoring that the issue was caused by the absence of a service layer. I had wanted the nullable result value to be clear but didn’t know where it should have been handled (Step 2 PR).

src/main/kotlin/ecommerce/product/ProductResponse.kt
miooo0o: To be completely honest, `data class` still feels like a `struct` to me. Even after reading the documentation, I sometimes unconsciously think of it that way. 
I tried to address this as much as possible in **Step 2.**

src/main/kotlin/ecommerce/product/Product.kt
miooo0o: That makes sense! I still find it challenging to think in terms of higher-level layers.
What do you think about using a mapper like this?

```kotlin
object ProductMapper {
    fun toResponse(product: Product): ProductResponse {
        return ProductResponse(
            product.id,
            product.name,
            product.price.toPlainString(),
            product.imageUrl,
        )
    }

    fun toEntity(
        request: ProductRequest,
        id: Long,
    ): Product {
        return Product(
            id = id,
            name = request.name,
            price = request.price,
            imageUrl = request.imageUrl,
        )
    }
}
```

Do you think this also violates the domain boundary?
In Step 2, I used this approach. If this is also considered problematic, I need to study more about how responses are typically handled...!

src/main/kotlin/ecommerce/product/Product.kt
miooo0o: https://github.com/herotechcourse/spring-ecommerce-product/pull/15#discussion_r2237971716 😿

src/main/kotlin/ecommerce/product/ProductRequest.kt
miooo0o: https://github.com/herotechcourse/spring-ecommerce-product/pull/15#discussion_r2237971716 😿

src/main/kotlin/ecommerce/product/ProductRequest.kt
miooo0o: https://github.com/herotechcourse/spring-ecommerce-product/pull/15#discussion_r2237976248

I know extensions aren’t generally recommended. At the time, my idea was to make entity conversion straightforward while leaving a clear point to revisit and refactor later.

Looking back, maintaining a proper feature list would’ve been a much better approach....!

src/main/kotlin/ecommerce/repository/ProductStore.kt
miooo0o: This method doesn’t actually generate the `id`. it inserts a record and retrieves the `generated id` using a `KeyHolder`.

The method name wasn’t clear, so I’ve renamed it to `insertAndReturnKey` for better readability.

src/test/kotlin/ecommerce/TextFixture.kt
miooo0o: I tried this after today’s commit, but it didn’t work as expected, so the current solution still feels a bit awkward. Since validation has been added to `ProductRequest`, `DummyRequest` isn’t really useful anymore. I may need to revisit and rework this in step 2 of the refactor if necessary. I’d love to hear your thoughts.

src/test/kotlin/ecommerce/TextFixture.kt
miooo0o: I don’t think it’s appropriate anymore.

I’m not sure how much of this remains in the Step 2 commit, but before I knew about proper test utilities (_may knowledge save your time_), I created test helpers to support E2E testing in `Fixture`, and some other file. It became much less readable. The test functions got shorter, but it became harder to understand what the tests were actually doing, making them inconvenient to work with.

From that experience, I learned that adding `assertions` inside a `Fixture` doesn’t make sense. It only makes tests more ambiguous and harder to maintain. It was a bit of a time sink, but I gained a valuable lesson from it.

PR #16, opened 2025-07-17
[spring-ecommerce-product - Step1] Submit knopgm's mission

knopgm:
# Project - Spring ecommerce-Product:

This application provides an API Product management UI toll, where users can create, retrieve, update and delete products from a data base, which are the basic CRUD operations fro any application.
It's implementation was on top of Spring Boot framework with Kotlin, H2 as a in-memory-database and a little of JavaScript to manipulation of the DOM.

# Features:

- Product -> a model class built to hold the products properties, methods and all that is required to be classified as a product
- ProductController -> responsible for store all CRUD HTTP API methods.
- ProductRepository -> stores all methods to manage data persistence and retrieval in database and uses JdbcTemplate to run SQL queries.
- ProductControllerTest -> to ensure that all CRUD methods are properly accessing data and answering users.

## Reflections:

Since this lesson had a lot of builded tools I feel like I had just a grasp of what this content really is. Nevertheless, with research, colleagues discussions and coaches guide, I manage to deliver this PR. My question is: How deep, as a beginner, should I go to understand each of this tools? Is there one that is more important than the other where I can focus for the benefit of the next tasks?
Another concern that I have in my mind is about the whole ecommerce architecture of the project: Since this project is going to have many feature implementations, do you have a better suggestion in how should I start the implementation of each main directory of it's features? Has Delivery Hero a good pattern to be used as example here?

Thank you for your time and reflections! 😄 🙏


src/main/kotlin/ecommerce/product/Product.kt
```diff
@@ -0,0 +1,18 @@
+package ecommerce.product
+
+class Product(var id: Long? = null, var name: String, var price: Double, var imageUrl: String?) {
+    fun update(newProduct: Product) {
+        this.name = newProduct.name
+        this.price = newProduct.price
+        this.imageUrl = newProduct.imageUrl
+    }
```
Hyeon9mak: In the context of entities, equality is determined based on the ID.

For example, if you have

- id = 1, name = "red apple"

On the other hand

```kotlin

        other as Product

        return id == other.id
    }

        return id?.hashCode() ?: 0
    }
```

README.md
```diff
@@ -1 +1,24 @@
...
+- [x] implement `update`
+  - if a product map is empty, return notFound()
+- [x] implement `delete`
+  - if a product map is empty, return notFound()
+  - if cannot find a product with id, return notFound()
+
+### Test: Product Controller
+- [x] refactor test names
+- [x] positive tests
+- [x] negative tests
```
Hyeon9mak, line 24: 👍 nice habit

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.PutMapping
+import org.springframework.web.bind.annotation.RequestBody
+import java.net.URI
+import java.util.concurrent.atomic.AtomicLong
+
+@Controller
+class ProductController(private val repository: ProductRepository) {
```
Hyeon9mak: What's different between `@RestController` & `@Controller` ?

and u can also append this annotation `@RequestMapping("/api/products")` 😉
what is it's role?

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,54 @@
...
+    @PutMapping("/api/products/{id}")
+    fun update(
+        @RequestBody newProduct: Product,
+        @PathVariable id: Long,
+    ): ResponseEntity<Product> {
+        val product = repository[id] ?: return ResponseEntity.notFound().build()
+        product.update(newProduct)
+        repository.updateById(id, product)
+        return ResponseEntity.ok().body(product)
+    }
```
Hyeon9mak: nice choice POST, GET, PUT, DELETE methods! 💯

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @DeleteMapping("/api/products/{id}")
+    fun delete(
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        repository.deleteById(id) ?: return ResponseEntity.notFound().build()
+        return ResponseEntity.noContent().build()
+    }
```
Hyeon9mak: 204 status code choice 👍

src/main/kotlin/ecommerce/repository/ProductRepository.kt
```diff
@@ -0,0 +1,72 @@
...
+            """.trimIndent()
+
+        return jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, id).takeIf { it == 1 }
+    }
+
+    fun deleteById(id: Long): Int? {
+        if (isEmptyOrNull()) return null
+        val sql = "DELETE FROM products WHERE id = ?"
+        return jdbcTemplate.update(sql, id).takeIf { it == 1 }
+    }
```
Hyeon9mak: how do we know when update or delete DML occur error?

src/main/resources/schema.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE products
+(
+    id        BIGINT PRIMARY KEY AUTO_INCREMENT,
+    name      VARCHAR(255),
+    price DOUBLE,
+    image_url TEXT
```
Hyeon9mak: name has length limit!

I think adding validation logic to the Product, might make the code more robust and easier to test. U think so too, right? 😄

Hyeon9mak (changes requested):
Hi Gabi, welcome to Spring mission!
I'm so sorry for late. 🙏 

You’ve implemented everything great!

With the Spring mission, I hope you won’t lose sight of the things you learned during the Blackjack mission.

Feel free to ask if you have any questions! 😄

Hyeon9mak:
> Since this lesson had a lot of builded tools I feel like I had just a grasp of what this content really is. Nevertheless, with research, colleagues discussions and coaches guide, I manage to deliver this PR. My question is: How deep, as a beginner, should I go to understand each of this tools? Is there one that is more important than the other where I can focus for the benefit of the next tasks?
Another concern that I have in my mind is about the whole ecommerce architecture of the project: Since this project is going to have many feature implementations, do you have a better suggestion in how should I start the implementation of each main directory of it's features? Has Delivery Hero a good pattern to be used as example here?

Of course, it's always good to invest time in learning tools thoroughly.
But just like we don't try to memorize everything in an encyclopedia, it's best to focus on learning the tools that are immediately relevant to your needs.
It might feel overwhelming now, with so many new concepts and tools to grasp—but over time, you'll realize that most of them follow similar patterns.
Take your time and learn them one by one.
For now, I recommend focusing especially on understanding the structure of Spring MVC.

Regardless of whether it's an ecommerce system or not, the directory structure and implementation style of any project should ultimately be designed to make the project easy to maintain.
And that’s something you and your team—including Gabi—can agree on together.
You can simply ask, “Would it be easier for us if we structured it this way?” and go ahead with that direction.

Throughout your time in the HeroTechCourse, you'll get the chance to try different approaches across various projects.
Through these experiences, you’ll come to understand the pros and cons of each, and eventually you'll know how to apply the right solution to the right context.

Just give it a try—there’s no need to overthink it right now!

src/main/kotlin/ecommerce/product/ProductController.kt
knopgm: @Controller: is normally used for traditional web apps that return views (HTML). It requires @ResponseBody on methods to return raw data.
@RestController: is used for APIs that return JSON and it automatically adds @ResponseBody to all methods.
@RequestMapping: sets a base path for all endpoints.

Thank you so much for asking! These annotations are making more sense in my code, indeed. 😄

src/main/kotlin/ecommerce/product/ProductController.kt
knopgm: PUT updates the full resource while PATCH updates only a field of the resource.
Amazing tip! I managed to apply the Patch instead of Put. 
Thank you! 🙏

src/main/kotlin/ecommerce/repository/ProductRepository.kt
knopgm: Good question! 
What do you think about this implementation for update:
 return try {
            jdbcTemplate.update(sql, product.name, product.price, product.imageUrl, id)
                .takeIf { it == 1 }
        } catch (ex: DataAccessException) {
            throw RuntimeException("Failed to update product with id $id", ex)
        }

and this for delete:
 return try {
            jdbcTemplate.update(sql, id).takeIf { it == 1 }
        } catch (ex: DataAccessException) {
            throw RuntimeException("Failed to update product with id $id", ex)
        }

?? 
I'm not sure how to simulate this error to test this catch... do you have any suggestions here? 
I'm going to push this refactoring, but I will study about validations and try to better understand this one.
If I did something wrong, you can always ask me to do again in another review, right?
Thank you for your patience! 🙏

src/main/resources/schema.sql
knopgm: I also think validations are great! 😄 
what do you think about something like this inside my ProductController:
@PostMapping("")
fun create(@RequestBody product: Product): ResponseEntity<Any> {
    // validation
    if (product.name.isNullOrBlank()) {
        return ResponseEntity.badRequest().body("Name cannot be blank")
    }

    if (product.name!!.length > 255) {
        return ResponseEntity.badRequest().body("Name must be at most 255 characters")
    }

    // current rest of the code
    val id = index.getAndIncrement()
    val newProduct = Product.toEntity(product, id)
    repository.insert(id, newProduct)
    return ResponseEntity.created(URI.create("/products/" + newProduct.id)).body(product)
}

I also already start to build a ProductService for the Step2 of this mission that is all about validations, so I'm not sure what is the difference between all this validations and which are essencial for the code. Maybe all of them, but I would like to know what do you think about this different validations and where can I look into to better distinguish their priorities. Thank you so much in advance! 🙏

src/main/kotlin/ecommerce/repository/ProductRepository.kt
Hyeon9mak: That's a great thought!
Whether the number of affected rows is 1 is a concern of the persistence layer, not the domain or application layer.
So it's best to immediately notify the failure. 💯 

The easiest case to test this would be when an ID is provided, but there's no matching data. 😄

src/main/resources/schema.sql
Hyeon9mak: If validation logic in the controller layer, it makes hard to generate test code. 😱 

In previous missions, you’ve practiced asking the domain layer about business rules.

For example, in the Blackjack mission,
rather than getting the hands of both the `Player` and the `Dealer`, summing the `Score`, and then making a judgment yourself, you would throw the `Player` to the `Dealer` and say, "You decide who wins.".

Similarly, with the `Product` class that U created, you could apply the same approach through its constructor—asking the same kind of business question.
Or, you might consider creating a value class like `ProductName` to encapsulate that logic. 😄

knopgm: Here I added a init block that result in type mismatch with my PATCH method that was requiring my product to have a null safety. To complete this init block requirement I created a ProductDTO object that has the null safety for its arguments, which allowed me to use the PATCH and the init block.
I hope I didn't lost myself in types for the whole files and tests I had to refactor because of that! 🙏 
Thank you so much again! I'll ask for a new review now.

src/main/kotlin/ecommerce/product/Product.kt
```diff
@@ -0,0 +1,46 @@
...
+    var id: Long? = null,
+    var name: String,
+    var price: BigDecimal,
+    var imageUrl: String?
+) {
+    init {
+        require(name.isNotBlank()) { "Name cannot be blank" }
+        require(name.length <= 255) { "Name must be at most 255 characters" }
+        require( price > BigDecimal.ZERO) { "Price must be positive" }
+    }
```
Hyeon9mak, line 15: 😉 👍 
i think imageUrl also have limit!

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+@RequestMapping("/api/products")
+@RestController
+class ProductController(private val repository: ProductRepository) {
+    private val index = AtomicLong(1)
+
+    @PostMapping("")
+    fun create(
+        @RequestBody product: Product,
+    ): ResponseEntity<Product> {
+        val id = index.getAndIncrement()
```
Hyeon9mak, line 25: i think generate entity's id role is persistence layer's role.
how do u think?

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+        val newProduct = Product.toEntity(product, id)
+        repository.insert(id, newProduct)
+
+        return ResponseEntity.created(URI.create("/products/" + newProduct.id)).body(product)
+    }
+
+    @GetMapping("")
```
Hyeon9mak, line 32:
```suggestion
    @GetMapping
```
😉

src/main/kotlin/ecommerce/product/ProductController.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @DeleteMapping("/{id}")
+    fun delete(
+        @PathVariable id: Long,
+    ): ResponseEntity<Void> {
+        repository.deleteById(id) ?: return ResponseEntity.notFound().build()
+        return ResponseEntity.noContent().build()
+    }
+}
```
Hyeon9mak, line 57: https://medium.com/@hyeon9mak/about-no-newline-at-end-of-file-warning-on-github-b42d14dccc93 😉

src/main/kotlin/ecommerce/product/ProdutDTO.kt
```diff
@@ -0,0 +1,15 @@
...
+    val name: String?,
+    val price: BigDecimal?,
+    val imageUrl: String?
+)
+
+fun ProductDTO.toProduct(): Product {
+    requireNotNull(name) { "Name is required" }
+    requireNotNull(price) { "Price is required" }
+    return Product(name = name, price = price, imageUrl = imageUrl)
+}
```
Hyeon9mak, line 15: not used. and Kotlin already gave null-safe data type. 🤔 

```kotlin
data class ProductDTO(
    val name: String,
)
```

src/main/kotlin/ecommerce/product/ProdutDTO.kt
```diff
@@ -0,0 +1,15 @@
+package ecommerce.product
+
+import java.math.BigDecimal
+
+data class ProductDTO(
+    val name: String?,
+    val price: BigDecimal?,
+    val imageUrl: String?
+)
```
Hyeon9mak, line 9: Why did u make `ProductDTO`? (just a question. feel free 😄)

Hyeon9mak (approved):
Hi Gabi! Looks good at reflect feedback! 👍 

U almost complete the mission requirements!
I left some simple comments.
it’s totally fine to go over them while working on the next step. 😄 

If you have any further questions, feel free to ask anytime!

src/main/kotlin/ecommerce/product/ProdutDTO.kt
knopgm: True. I just thought that it may be needed for the next steps of the Product development, so I leave it there to check... But maybe it is only my beginner inexperience talking ahead of me. 😅

src/main/kotlin/ecommerce/product/ProdutDTO.kt
knopgm: In my research, it was a good option to make a shallow copy of the Product Entity and modify it's types to better accept the PATCH method with type safety, but it can also: 

- protect the domain from direct interaction with the API
- make easier to test because it's easier to mock DTOs for unit testing.
- and for future improvements it makes easiar to versioning not to affect the client

But I also read some downsides to it, like:
- it can add complexity where it isn't necessary (small project)
- can be repetitive and difficult maintenance

With all that in mind I'm not sure if I applied the right decision for my PATCH problem. but since I'm using only for that, maybe it is not such a complexity to worry about right?
What is your opinion about it? Do you have a better recommendation to use PATCH without adding more complexity?

Thank you in advance! 🙏

src/main/kotlin/ecommerce/product/ProdutDTO.kt
knopgm: I improve this function to this: 
```
fun validate(): ProductDTO {
        require(!name.isNullOrBlank()) { "Name must not be blank." }
        require(price != null && price > BigDecimal.ZERO) { "Price must be greater than 0." }
        require(!imageUrl.isNullOrBlank() && imageUrl.startsWith("http")) { "Invalid image URL." }
        return this
 }
```

This way I can validate the inputs on my PATCH method! 😄 
Is that a good path for this function?

src/main/kotlin/ecommerce/product/ProductController.kt
knopgm: Done!

PR #18, opened 2025-07-25
[spring-ecommerce-product - Step-2] Submit sebasnadu's mission

SebasNadu:
Hey Daeun!👋

Hope you’re doing well. Here’s my Step 2 submission! I tried to build something cool that brings together everything we’ve been learning in the course.
Hope you enjoy checking it out,  and that my code’s easy to follow.
Looking forward to your feedback!✌️

## 🚀 Feature Complete: Product, Member, Cart & Admin Stats API Integration

This PR delivers full support for the core functional requirements of the `spring-ecommerce-productDTO` backend, implementing product validation, secure member authentication, cart management, and admin-facing statistics endpoints.

---

#### 🛒 **Product CRUD + Validation**
- Full support for creating, updating, retrieving, and deleting products.
- Enforced business rules:
  - Name ≤ 15 characters.
  - Allowed characters: alphanumeric + `()[]+-&/_`.
  - Must be unique.
  - Price must be > 0.
  - `imageUrl` must start with `http://` or `https://`.
- Validation failures return structured error responses for client-side usability.

#### 👤 **Member Registration & Login with JWT**
- `POST /api/members/register` — Register a new user and return a token.
- `POST /api/members/login` — Authenticate existing user and issue JWT.
- Stateless authentication via `Authorization: Bearer <token>`.
- Use of interceptor to ask for the token on the request

#### 🛍️ **Cart Management for Authenticated Users**
- `GET /api/cart` — View user’s current cart items.
- `POST /api/cart` — Add a product to the cart.
- `DELETE /api/cart/{productId}` — Remove a product from the cart.
- Token required for all cart operations.

#### 📊 **Admin-Only Statistics**
- `GET /api/admin/stats/top-products` — Top 5 most-added products to carts in the last 30 days. Ordered by frequency, then recency.
  - Response includes: product name, added count, last added time.
- `GET /api/admin/stats/active-members` — Members who added at least one item in the last 7 days.
  - Response includes: member ID, name, and email.
- Restricted to users with `ADMIN` role via custom annotations.

---

### 🔐 Security & Access Control
- Added role-based access control using JWT tokens.
- Custom annotations:
  - `@LoginMember` — Injects authenticated user using `resolvers`.
  - `@CheckAdminOnly` — Restricts endpoint access to admin users.
  - `@IgnoreCheckLogin` — Allow the access of this endpoint without token (easy to customize when different endpoints has different methods and you want to require token in one of them. e.g. /api/products for get public, for post, put and delete restricted by token and only admin).

---

### 💡 Notes
- Introduced domain models: `Member`, `CartItem`, and `Product`, all mapped to corresponding DTOs.
- Service, repository, controller, infrastructure and config layers follow clear separation of concerns.
- Comprehensive exception handling for REST APIs, including validation and authorization errors with a global advice controller.
- E2E tests
- Integration test for the services.
- possible TODO's: Implement some views for the new features.


README.md
```diff
@@ -57,4 +57,60 @@
   - [x] Add findAllPaginated() to the service and repository
   - [x] Add countAll() to the repository, returning the total number of products
   - [x] Add a pagination bar to the product list
-  - [x] Add a link to the pagination bar and a simple logic to handle the page number
\ No newline at end of file
+  - [x] Add a link to the pagination bar and a simple logic to handle the page number
+### Step 2.1
```
da-nyee, line 61: So many requirements 😮

build.gradle.kts
```diff
@@ -25,8 +25,13 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.h2database:h2:2.2.224")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
     implementation("org.springframework.boot:spring-boot-starter-validation")
+    implementation("io.jsonwebtoken:jjwt-api:0.12.6")
+//    compileOnly("io.jsonwebtoken:jjwt-api:0.12.6")
```
da-nyee: You might consider removing this comment if the dependency isn't needed 👀

build.gradle.kts
```diff
@@ -25,8 +25,13 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+    implementation("com.h2database:h2:2.2.224")
```
da-nyee: I think it’s better to define it as `runtimeOnly`, since the database is only needed at runtime.

src/main/kotlin/ecommerce/controller/CartItemController.kt
```diff
@@ -0,0 +1,44 @@
...
+    @GetMapping
+    fun getCartItemsWithProducts(
+        @LoginMember member: MemberDTO,
+    ): ResponseEntity<List<CartItemResponseDTO>> {
+        val cartItems = cartItemService.findByMember(member.id!!)
+        return ResponseEntity.ok().body(cartItems)
+    }
+
+    @PostMapping
+    fun addOrUpdateCartItem(
```
da-nyee, line 28: How about splitting this into a `POST` API and a `PUT` or `PATCH` API?
It seems like this one is handling too many responsibilities.

src/main/kotlin/ecommerce/controller/ProductViewController.kt
```diff
@@ -12,23 +15,43 @@ import org.springframework.web.bind.annotation.RequestParam
...
+    fun createProduct(
+        @Valid productDTO: ProductDTO,
+        bindingResult: BindingResult,
+        model: Model,
+        @RequestParam(required = false) pageNumber: Int = 1,
+        @RequestParam(required = false) pageSize: Int = 10,
+    ): String {
+        if (bindingResult.hasErrors()) {
+            val (products, totalCount) = productService.findAllPaginated(pageNumber, pageSize)
+            val totalPages = (totalCount + pageSize - 1) / pageSize
```
da-nyee: It looks like there’s some duplicated code — could you simplify it?

src/main/kotlin/ecommerce/model/ProductDTO.kt
```diff
@@ -1,9 +1,24 @@
...
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Positive
+import jakarta.validation.constraints.Size
+
 data class ProductDTO(
     var id: Long? = null,
+    @field:NotBlank
+    @field:NotNull
+    @field:Size(min = 1, max = 15, message = "The product name must contain between 1 and 15 characters")
+    @field:Pattern(regexp = "^[a-zA-Z0-9 ()\\[\\]+\\-&/_]*$", message = "Invalid characters in product name.")
```
da-nyee: Regex 👍

src/main/kotlin/ecommerce/model/ProductDTO.kt
```diff
@@ -1,9 +1,24 @@
...
     var name: String,
+    @field:NotNull
+    @field:Positive
```
da-nyee: Could you add a message for this?

image

There's one test failure, and the message is returned in Korean.
I believe that's because my laptop's default language is set to Korean.
It might be better to clarify the message so that the test works correctly in any environment.

src/main/kotlin/ecommerce/model/ActiveMemberDTO.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.model
+
+data class ActiveMemberDTO(
+    val id: Long,
+    val email: String,
```
da-nyee, line 13:
```suggestion
    @field:Email
    val email: String,
```

Just a small suggestion 😄

src/main/kotlin/ecommerce/annotation/LoginMember.kt
```diff
@@ -0,0 +1,5 @@
+package ecommerce.annotation
+
+@Target(AnnotationTarget.VALUE_PARAMETER)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class LoginMember
```
da-nyee, line 5: Custom annotation 👍

src/main/kotlin/ecommerce/config/interceptor/CheckLoginInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+import ecommerce.infrastructure.JwtTokenProvider
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.stereotype.Component
+import org.springframework.web.method.HandlerMethod
+import org.springframework.web.servlet.HandlerInterceptor
+
+@Component
+class CheckLoginInterceptor(private val jwtTokenProvider: JwtTokenProvider) : HandlerInterceptor {
+    override fun preHandle(
```
da-nyee, line 15: It could be helpful to understand how `DispatcherServlet` works 👀

src/main/kotlin/ecommerce/mappers/CartItemMapper.kt
```diff
@@ -0,0 +1,9 @@
+package ecommerce.mappers
+
+import ecommerce.entities.CartItem
+import ecommerce.model.CartItemResponseDTO
+import ecommerce.model.ProductDTO
+
+fun CartItem.toResponseDto(productDTO: ProductDTO) = CartItemResponseDTO(id!!, memberId, productDTO, quantity, addedAt!!)
+
+fun CartItemResponseDTO.toEntity() = CartItem(id, memberId, product.id!!, quantity, addedAt)
```
da-nyee: It seems like this one isn't used 👀

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
```diff
@@ -0,0 +1,28 @@
+package ecommerce.infrastructure
+
+import jakarta.servlet.http.HttpServletRequest
+import org.springframework.stereotype.Component
+
+@Component
+class AuthorizationExtractor {
+    fun extractToken(request: HttpServletRequest): String {
```
da-nyee, line 9: How about extracting some of the code into separate methods? It looks a bit lengthy.

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
```diff
@@ -0,0 +1,28 @@
...
+                    authHeaderValue = authHeaderValue.substring(0, commaIndex)
+                }
+                return authHeaderValue
+            }
+        }
+        return ""
+    }
+
+    companion object {
+        private const val AUTHORIZATION = "Authorization"
```
da-nyee, line 30: What's the difference between `Authentication` and `Authorization`? I'd like to hear your take on it.

src/main/kotlin/ecommerce/config/resolvers/LoginMemberArgumentResolver.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.web.bind.support.WebDataBinderFactory
+import org.springframework.web.context.request.NativeWebRequest
+import org.springframework.web.context.request.RequestAttributes
+import org.springframework.web.method.support.HandlerMethodArgumentResolver
+import org.springframework.web.method.support.ModelAndViewContainer
+
+@Component
+class LoginMemberArgumentResolver(
+    private val memberService: MemberService,
+) : HandlerMethodArgumentResolver {
```
da-nyee, line 17: Good use of `HandlerInterceptor` and `HandlerMethodArgumentResolver` 👍

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
```diff
@@ -0,0 +1,199 @@
...
+import ecommerce.model.ActiveMemberDTO
+import ecommerce.model.TopProductDTO
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.jdbc.core.simple.SimpleJdbcInsert
+import org.springframework.stereotype.Repository
+import java.sql.ResultSet
+
+@Repository
+class CartItemRepositoryImpl(private val jdbc: JdbcTemplate) : CartItemRepository {
```
da-nyee, line 14: Does it really need the interface–implementation structure?
It feels a bit verbose, since there's only one implementation per interface.

src/main/kotlin/ecommerce/services/AdminServiceImpl.kt
```diff
@@ -0,0 +1,17 @@
+package ecommerce.services
+
+import ecommerce.model.ActiveMemberDTO
+import ecommerce.model.TopProductDTO
+import ecommerce.repositories.CartItemRepository
+import org.springframework.stereotype.Service
+
+@Service
+class AdminServiceImpl(private val cartItemRepository: CartItemRepository) : AdminService {
```
da-nyee, line 9: Same as the comment in `CartItemRepositoryImpl`.

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
```diff
@@ -0,0 +1,199 @@
...
+            """
+            SELECT p.name,
+                   COUNT(*) AS added_count,
+                   MAX(c.added_at) AS most_recent_added_at
+            FROM cart_item c
+            JOIN product p ON c.product_id = p.id
+            WHERE c.added_at >=  DATEADD('DAY', -30, CURRENT_TIMESTAMP)
+            GROUP BY c.product_id, p.name
+            ORDER BY added_count DESC, most_recent_added_at DESC
+            LIMIT 5
```
da-nyee, line 173: 👍

da-nyee: If you were to add indexes on `cart_item` and `product` to improve performance, how would you do it?

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
```diff
@@ -0,0 +1,199 @@
...
+        return jdbc.update(sql, cartItem.productId, cartItem.memberId) > 0
+    }
+
+    override fun findDistinctMembersWithCartActivityInLast7Days(): List<ActiveMemberDTO> {
+        val sql =
+            """
+            SELECT DISTINCT m.id, m.email
+            FROM cart_item c
+            JOIN member m ON c.member_id = m.id
+            WHERE c.added_at >= DATEADD('DAY', -7, CURRENT_TIMESTAMP)
```
da-nyee: image

I think you might have missed one point — could you take another look?

src/test/kotlin/ecommerce/endtoend/ProductE2ETest.kt
```diff
@@ -0,0 +1,263 @@
...
+        assertThat(response.statusCode()).isEqualTo(HttpStatus.NOT_FOUND.value())
+    }
+
+    @Test
+    fun createProduct() {
+        val newProductDTO = ProductDTO(name = "Monitor", price = 150.0, imageUrl = "https://example.com/monitor.jpg")
+
+        val response =
+            RestAssured.given().log().all()
+                .auth().oauth2(token) // <-- add this line to send the token
```
da-nyee: Is this comment really necessary? 👀

src/test/kotlin/ecommerce/integration/MemberServiceTest.kt
```diff
@@ -0,0 +1,138 @@
...
+                memberService.save(MemberDTO(email = "exists@test.com", password = "new"))
+            }
+
+        assertThat(ex.message).contains("already exists")
+    }
+
+    @Test
+    fun `validateEmailUniqueness should pass if email doesn't exist`() {
+        assertThat(memberService.validateEmailUniqueness("unique@test.com")).isEqualTo(Unit)
+        // no exception = pass
```
da-nyee: I’d recommend removing this comment if it’s not needed.

src/test/kotlin/ecommerce/integration/ProductServiceJDBCTest.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.autoconfigure.jdbc.AutoConfigureTestDatabase
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.dao.DataAccessException
+import org.springframework.test.annotation.DirtiesContext
+import org.springframework.transaction.annotation.Transactional
+
+@SpringBootTest
+@AutoConfigureTestDatabase
+@Transactional
```
da-nyee, line 18: It behaves differently when we use `@Transactional` in tests. Have you noticed that?

src/test/kotlin/ecommerce/endtoend/AdminE2ETest.kt
```diff
@@ -0,0 +1,119 @@
...
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.http.MediaType
+import org.springframework.test.annotation.DirtiesContext
+import java.time.LocalDateTime
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+class AdminE2ETest {
```
da-nyee, line 19: Have you had any difficulties writing E2E tests?

da-nyee (changes requested):
Hi Sebas! 🙌
You're doing really great 💯
I’ve gone through it, and it looks like you’ve met all the requirements.

I left a few comments — feel free to check them out.
If you have any questions, just let me know 🙇‍♀️
And if there's anything else you'd like me to take a closer look at, I’m happy to help ☺️

build.gradle.kts
SebasNadu: Yes!, that makes total sense, H2 is only needed at runtime, and I’m not using it anywhere during compilation. Thanks for pointing that out!

src/main/kotlin/ecommerce/controller/CartItemController.kt
SebasNadu: I actually considered splitting it into POST and PATCH, but in this case, I decided to unify the logic in a single endpoint based on how I envision the client UI working.

When a user is on a product page and adds the item to their cart, if the product is already there, the API simply updates the quantity instead of creating a new entry. The key point is that the client uses the same button for both “add” and “update” actions. I prefer to delegate this logic to the backend rather than requiring the frontend to manage it, otherwise, the frontend would need to either store the current cart state (e.g., in a SPA) or make multiple requests to determine the correct operation.

Given the current project scope and the possible frontend implementation, I think it makes more sense for the server-side service to handle whether it’s effectively a POST or a PATCH. Also, when the quantity is set to 0, I don’t delete the item, I keep it for potential future use, like analytics or reorder history.

So instead of modeling the request as a strict “create vs update”, the endpoint reflects a single user intent: sync the cart with this product. This leads to fewer requests, a simpler frontend, and no need to expose internal cart item IDs to the client.

I’m open to revisiting this later if the requirements evolve, but for now I believe this design keeps things clean and cohesive. Let me know what you think!😉

src/main/kotlin/ecommerce/controller/ProductViewController.kt
SebasNadu: That's right, good catch! Thank you!

src/main/kotlin/ecommerce/model/ProductDTO.kt
SebasNadu: That's very interesting!, without a custom message, the default one depends on the system locale.  That's right, it can cause inconsistent behaviour in tests.
I’ll add explicit messages like this:

@field:NotNull(message = "Price must not be null")
@field:Positive(message = "Price must be greater than zero")

That should make the validation predictable across different environments, regardless of locale. Thank you!🎉

src/main/kotlin/ecommerce/config/interceptor/CheckLoginInterceptor.kt
SebasNadu: DispatcherServlet handles all incoming HTTP requests. It parses the request and delegates it to the appropriate controller (@Controller or @RestController) using HandlerMapping.

Before reaching the controller method, any registered interceptors run their preHandle() method. If preHandle() returns false, the request is aborted and won’t reach the controller. If it returns true, the controller processes the request.

After the controller returns, but before the view is rendered, postHandle() is called. This is a good place to modify the ModelAndView.

Finally, once view rendering is complete and just before the response is sent, afterCompletion() runs. This can be used for tasks like logging or resource cleanup.

That’s how I understand the flow, feel free to correct me if I missed something.

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
SebasNadu: Authentication is about checking who you are, basically verifying your identity. In this project, that happens during login when we create the token.

Authorization comes after that, it checks what you’re allowed to do. For example, accessing protected routes like /api/admin. In our case, that’s handled by the CheckRoleInterceptor, which ensures you have the right role.

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
SebasNadu: Yeah, it is a bit verbose and definitely adds some overhead when adding or changing methods. But I chose this approach to follow good practices and make the project easier to scale and test in the future. Using an interface lets me mock the repository for unit testing, and makes it easier to switch implementations down the road.

For example, I think next week we’ll start working with JPA. When that happens, I’ll likely need to migrate the service layer to use a JPA-based repository instead of the current JDBC one. Having an interface in place should make that transition smoother, I hope🤞🤞.

src/main/kotlin/ecommerce/services/AdminServiceImpl.kt
SebasNadu: Same reason ☝

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
SebasNadu: Yes, you’re right, adding more indexes can improve these specific queries. Here are the indexes I think would be most helpful for the current state:
- CREATE INDEX idx_cart_item_member_id ON cart_item(member_id);
Supports queries filtering by member, like findByMember.
- CREATE INDEX idx_cart_item_product_member ON cart_item(product_id, member_id);
Optimizes queries filtering by both product and member IDs, such as existsByProductAndMember and update.
- CREATE INDEX idx_cart_item_added_at ON cart_item(added_at);
Helps time-based queries like findDistinctMembersWithCartActivityInLast7Days.
- CREATE INDEX idx_cart_item_product_added ON cart_item(product_id, added_at);
Supports filtering, grouping, and joins in queries like findTop5ProductsAddedInLast30Days.

Let me know if you have another indexes ideas that improve my queries 🤔😉

src/test/kotlin/ecommerce/integration/ProductServiceJDBCTest.kt
SebasNadu: Yes, you’re right. Honestly, at the time I didn’t fully understand what was going on, I was just trying to get the tests to run correctly. But after reviewing the setup more carefully, I realized that @Transactional already rolls back the changes after each test, which gives proper isolation.

Given that, using @DirtiesContext adds unnecessary overhead without any real benefit in this context. So I’ve removed it.

Also, the @BeforeEach that clears the entire database with deleteAll() was overkill. Since the transaction rollback already resets the state, I refactored the tests to remove that part too.

Now the tests run faster and are more focused. Thanks for pointing this out, it really reminded me how important it is to understand your tools🙌!

src/test/kotlin/ecommerce/endtoend/AdminE2ETest.kt
SebasNadu: At first, not really, I actually started with E2E tests, and they felt pretty straightforward. I struggled more with integration tests in the beginning.

But now, as I’ve been learning more about testing in Spring, I’ve started to question a few things. For example, when should I use @DirtiesContext versus @Transactional? Early on, I was using @DirtiesContext by default without really thinking about whether it was necessary. Now I understand that it can be quite heavy and often avoidable if you’re using proper transactional rollbacks.

Another thing that clicked recently: if you use your production database for testing, the tests become less predictable over time as the data changes. That made me appreciate using a dummy database like H2 for testing, you get consistent, predefined data and don’t have to worry about the current state of the DB.

I still have a lot to learn and practice when it comes to testing in Spring, but I’m improving every day. Thanks for helping me along the way!🙏🙏

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
SebasNadu: Ouch, you’re right, I totally missed that part 😅. Thanks for pointing it out!

SebasNadu:
Hi Daeun!
I implemented your suggestion: Improve validation, optimize queries, and extend member data (name)

What’s new:
	•	➕ Added name field to Member and ActiveMemberDTO, updated services and tests accordingly.
	•	✅ Enhanced validation with custom messages in all DTOs for better user feedback.
	•	🔍 Refactored token extraction logic for cleaner authorization handling.
	•	🧠 Optimized SQL query for active members and added indexes to improve performance.
	•	🧪 Added unit tests for JwtTokenProvider and AuthorizationExtractor in the infrastructure layer.
	•	🧹 Cleaned up tests: removed unnecessary DirtiesContext usage and improved assertion clarity.
	•	✂️ Removed unused imports from ProductE2ETest.

Thanks for the feedback! It really helped me understand some concepts better and keep growing as a developer 🫰🙂

src/main/kotlin/ecommerce/controller/ProductViewController.kt
```diff
@@ -12,24 +15,47 @@ import org.springframework.web.bind.annotation.RequestParam
...
+            loadProductList(model, pageNumber, pageSize)
+            model.addAttribute("productDTO", productDTO)
+            model.addAttribute("hasErrors", bindingResult.hasErrors())
+            return "product-list"
+        }
         productService.save(productDTO)
         return "redirect:/"
     }
+
+    private fun loadProductList(
```
da-nyee, line 49: 👍

src/main/kotlin/ecommerce/config/interceptor/CheckLoginInterceptor.kt
da-nyee: You nailed it 💯

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
```diff
@@ -0,0 +1,33 @@
...
+        val headerValue = extractAuthorizationHeader(request.getHeaders(AUTHORIZATION)) ?: return ""
+        var authHeaderValue = headerValue.substring(BEARER_TYPE.length).trim()
+        val commaIndex = authHeaderValue.indexOf(',')
+        if (commaIndex > 0) {
+            authHeaderValue = authHeaderValue.substring(0, commaIndex)
+        }
+        return authHeaderValue
+    }
+
+    private fun extractAuthorizationHeader(headers: Enumeration<String>): String? {
```
da-nyee, line 19: 👍

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
da-nyee: Perfect 💯

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
da-nyee: Great, that makes sense 👍
I see why you chose this approach 😄

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
```diff
@@ -0,0 +1,189 @@
...
+
+    override fun deleteByProduct(cartItem: CartItem): Boolean {
+        val sql = "DELETE FROM cart_item WHERE product_id = ? And member_id = ?"
+        return jdbc.update(sql, cartItem.productId, cartItem.memberId) > 0
+    }
+
+    override fun findDistinctMembersWithCartActivityInLast7Days(): List<ActiveMemberDTO> {
+        val sql =
+            """
+            SELECT DISTINCT m.id, m.email, m.name
```
da-nyee, line 147: 👍

src/main/kotlin/ecommerce/controller/CartItemController.kt
da-nyee: Okay, it really makes sense 👍
This design works well given the current context.

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
da-nyee: You guessed well 👍 
But I’d recommend considering a couple of different indexes.

Have you heard of `covering indexes`?
They can significantly improve query performance in some cases.

So here’s what I'd suggest:

```sql
cart_item (added_at, product_id)
product (id, name)
```

Also, have you heard of `EXPLAIN`?
It shows how the database executes a query — definitely worth checking out ☺️

src/test/kotlin/ecommerce/endtoend/AdminE2ETest.kt
da-nyee: You're learning and growing every day 👍
Stay motivated and keep going 💪

da-nyee (changes requested):
Hi Sebas! 🙌
You've been doing such great work — I can really see your growth 👏

It looks like the mission is ready to wrap up.
But I'll leave it as a request change for now, since I've added a few comments — please take a look.

If anything’s unclear, feel free to let me know.
Otherwise, once everything’s clear, just request a review again — I’ll go ahead and merge it ☺️

src/main/kotlin/ecommerce/repositories/CartItemRepositoryImpl.kt
SebasNadu: Wow, thank you! I checked with EXPLAIN and realized my indexes weren’t being used the way I expected. Your suggestion helped me understand how covering indexes actually work, super useful!. Appreciate the tip! 🙌

src/main/resources/schema.sql
```diff
@@ -1,8 +1,36 @@
...
+    quantity   INT       NOT NULL DEFAULT 1,
+    added_at   TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    UNIQUE (member_id, product_id),
+    CONSTRAINT fk_member FOREIGN KEY (member_id) REFERENCES MEMBER (id) ON DELETE CASCADE,
+    CONSTRAINT fk_product FOREIGN KEY (product_id) REFERENCES PRODUCT (id) ON DELETE CASCADE
+);
+
+CREATE INDEX idx_cart_item_cover ON cart_item(added_at, product_id);
+CREATE INDEX idx_product_cover ON product(id, name);
```
da-nyee, line 36: Thanks for applying the feedback 🙇‍♀️

da-nyee (approved):
Hi Sebas! 🙌

It's finally time to wrap up the mission — nice job 🎉
You've been doing really well, even though this one was a tough challenge.
I hope the feedback has been helpful for your growth and kept you motivated in your development journey.

If you have any further questions, feel free to reach out — always happy to help ☺️

PR #19, opened 2025-07-25
Step2

deniz-oezdemir:
H there! 👋

I hope you are doing well!

Just for context, I wanted to share a few things:
I originally over-engineered the project by pulling in Spring Security and hashing passwords. In hindsight, that made things more complex than they needed to be, especially because I didn’t fully understand how Spring Security automatically interacts with the rest of the program. I ended up reversing all of it. It wasn’t wasted time though as I learned quite a bit about security.
That said, it cost me most of Wednesday and put me a bit behind the schedule I had set for myself. So things got a little hectic towards the end, so while I tried to keep the code readable and clean, I’m sure there are areas that could be improved. I focused on getting this first PR done before the weekend rather than making everything perfect.

A couple of questions I would appreciate your input on:
- Do we want users to have a name field? I noticed this came up when building the recent active users feature. For now, for the admin statistics about recently active users I just use the part before the @ in the email. It works, but I can add a proper name field if that’s something we want.
- I added a list of future work items in the README. I tried to sort them by priority. I would love your thoughts on whether that order makes sense.

In general, the overall structure (DB → Entity → Repository → Service → Controller) was new to me. I tried to follow the general pattern, but I’m not sure if I applied it well across the project. Any feedback on that would be especially helpful.

Thanks so much for reviewing this! 

Have a great weekend 😊

All the best,
Deniz


deniz-oezdemir:
Closed because of conflicts. Will fix conflicts and open a new PR.

PR #21, opened 2025-07-25
[spring-ecommerce-product - Step-2] Submit deniz-oezdemir's mission

deniz-oezdemir:
Hi GuMak! 👋

I hope you are doing well!

Just for context, I wanted to share a few things: I originally over-engineered the project by pulling in Spring Security and hashing passwords. In hindsight, that made things more complex than they needed to be, especially because I didn’t fully understand how Spring Security automatically interacts with the rest of the program. I ended up reversing all of it. It wasn’t wasted time though as I learned quite a bit about security. That said, it cost me most of Wednesday and put me a bit behind the schedule I had set for myself. So things got a little hectic towards the end, so while I tried to keep the code readable and clean, I’m sure there are areas that could be improved. I focused on fulfilling the requirements within the deadline rather than making everything perfect.

A couple of questions I would appreciate your input on:
- Do we want users to have a name field? I noticed this came up when building the recent active users feature. For now, for the admin statistics about recently active users I just use the part before the @ in the email. It works, but I can add a proper name field if that’s something we want.
- I added a list of future work items in the README. I tried to sort them by priority. I would love your thoughts on whether that order makes sense.

In general, the overall structure (DB → Entity → Repository → Service → Controller) was new to me. I tried to follow the general pattern, but I’m not sure if I applied it well across the project. Any feedback on that would be especially helpful.

I also implemented the last open change requests from your review of [step1](https://github.com/herotechcourse/spring-ecommerce-product/pull/1).

Thanks so much for reviewing this!

Have a great weekend 😊

All the best,
Deniz


deniz-oezdemir:
After a few failed attempts at resolving the conflicts, I think I finally got it right this time 😄
Let me know if anything still looks off!

src/test/kotlin/ecommerce/dto/ProductRequestTest.kt
```diff
@@ -0,0 +1,113 @@
...
+            ProductRequest(
+                name = "Invalid@Name!",
+                price = 1.99,
+                imageUrl = "https://example.com/image.jpg",
+            )
+
+        val violations = validator.validate(product)
+
+        Assertions.assertThat(violations).hasSize(1)
+        Assertions.assertThat(violations.first().message).isEqualTo("Name contains invalid special characters.")
```
Hyeon9mak: nice test cases 👍

src/main/kotlin/ecommerce/controller/MemberController.kt
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/api/members")
+class MemberController(
+    val userService: UserService,
+    val jwtService: JwtService,
```
Hyeon9mak: what is the different of member and user?
why separate two word?

src/main/kotlin/ecommerce/controller/MemberController.kt
```diff
@@ -0,0 +1,48 @@
...
+    fun register(
+        @Valid @RequestBody request: RegisterRequest,
+    ): ResponseEntity<RegisterResponse> {
+        val role = request.role ?: "USER"
+
+        userService.register(request.email, request.password, role)
+        val token = jwtService.generateToken(request.email)
+        val response = RegisterResponse(token)
+        return ResponseEntity.ok(response)
+    }
```
Hyeon9mak: "if user no choice role, default role is `USER`".
I think it's not a HTTP protocol's logic. It's business logic.
How do u think?

src/main/kotlin/ecommerce/controller/MemberController.kt
```diff
@@ -0,0 +1,48 @@
...
+        val user =
+            userService.getByEmail(request.email)
+                ?: return ResponseEntity.status(403).build()
+
+        val passwordMatches = userService.checkPassword(user, request.password)
+        if (!passwordMatches) return ResponseEntity.status(403).build()
+
+        val token = jwtService.generateToken(user.email)
+        return ResponseEntity.ok(LoginResponse(token))
+    }
```
Hyeon9mak: Also, we learned about "how to send message to object"!
we can ask like this.

```kotlin
// MemberController.kt
userService.login(request.email, request.password)
```

```kotlin
// UserService.kt
user.checkPassword(password)
```

```kotlin
// User.kt
throw CustomException()
```

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @DeleteMapping("/{productId}")
+    fun removeItem(
+        @LoginMember user: User,
+        @PathVariable productId: Long,
+    ): ResponseEntity<Void> {
+        cartService.removeItem(user.id!!, productId)
+        return ResponseEntity.noContent().build()
+    }
```
Hyeon9mak: Come to think of it, the User object was originally designed to map 1:1 with the database table, but it’s also being used to carry around information about the logged-in user. That means the object now has at least two different reasons to change.

Moreover, some fields like the password aren't even directly used in the context of the logged-in user. Would it be worth considering separating them into different objects?

src/main/kotlin/ecommerce/config/AdminRoleInterceptor.kt
```diff
@@ -0,0 +1,45 @@
...
+        val token = authHeader.removePrefix("Bearer ").trim()
+        val email = jwtService.validateAndExtractEmail(token)
+
+        val user =
+            userService.getByEmail(email)
+                ?: throw ResponseStatusException(HttpStatus.UNAUTHORIZED, "User not found")
+
+        if (user.role != "ADMIN") {
+            throw ResponseStatusException(HttpStatus.UNAUTHORIZED, "Access denied: ADMIN role required")
+        }
```
Hyeon9mak: What's different of FORBIDDEN and UNAUTHORIZED?

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+class GlobalExceptionHandler {
+    @ExceptionHandler(MethodArgumentNotValidException::class)
+    fun handleValidationException(ex: MethodArgumentNotValidException): ResponseEntity<Map<String, String>> {
+        val errors =
+            ex.bindingResult
+                .fieldErrors
+                .associate { it.field to (it.defaultMessage ?: "Invalid value") }
+
+        return ResponseEntity.badRequest().body(errors)
+    }
```
Hyeon9mak, line 40: 😄 👍

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun handleResponseStatusException(ex: ResponseStatusException): ResponseEntity<Map<String, String>> {
+        val body = mapOf("error" to (ex.reason ?: "Unexpected error"))
+        return ResponseEntity(body, ex.statusCode)
+    }
+
+    @ExceptionHandler(RuntimeException::class)
+    fun handleRuntimeException(ex: RuntimeException): ResponseEntity<Map<String, String>> {
+        val body = mapOf("error" to (ex.message ?: "Unexpected error"))
+        return ResponseEntity(body, HttpStatus.BAD_REQUEST)
+    }
```
Hyeon9mak: `RuntimeException` is a very broad category—can we really assume that all such cases are due to user error and return a 400? If the issue is actually on the server side, wouldn't the user end up wasting time wondering, 'Did I do something wrong?'

src/main/kotlin/ecommerce/config/WebMvcConfig.kt
```diff
@@ -0,0 +1,31 @@
...
+                "/api/products",
+                "/api/members/register",
+                "/api/members/login",
+                "/debug/**",
+            )
+
+        registry.addInterceptor(adminRoleInterceptor)
+            .addPathPatterns("/admin/**")
+    }
+
```
Hyeon9mak: Similar to how adminRoleInterceptor clearly indicates the role it's targeting, maybe consider renaming jwtAuthInterceptor to reflect the specific role it handles—since JWT is just an authentication mechanism, not the role itself. 😄

src/main/kotlin/ecommerce/repository/CartItemHistoryRepository.kt
```diff
@@ -0,0 +1,61 @@
...
+            JOIN products p ON h.product_id = p.id
+            WHERE h.action = 'ADD'
+              AND h.created_at >= DATEADD('DAY', -30, CURRENT_TIMESTAMP)
+            GROUP BY p.id, p.name
+            ORDER BY add_count DESC, latest_added_at DESC
+            LIMIT 5
+            """.trimIndent()
+
+        return jdbc.queryForList(sql)
+    }
```
Hyeon9mak: The CartRepository was kind enough to return a Boolean based on the number of affected rows during update or delete operations, but this time the repository feels a bit too raw. It seems like two different design concepts are clashing under the single name 'Repository'. How about deciding which approach is better and unifying the design accordingly? 🤔

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.web.server.ResponseStatusException
+
+@Service
+class ProductService(private val productRepository: ProductRepository) {
+    fun validateUniqueName(name: String) {
+        if (productRepository.existsByName(name)) {
+            throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Product name must be unique.")
+        }
+    }
+
```
Hyeon9mak: 🤔 private?

src/main/kotlin/ecommerce/dto/ProductRequest.kt
```diff
@@ -0,0 +1,34 @@
...
+    )
+    val name: String,
+    @field:DecimalMin(value = "0.01", message = "Price must be greater than 0.")
+    val price: Double,
+    @field:NotBlank
+    @field:Pattern(
+        regexp = "^https?://.*",
+        message = "Image URL must start with http:// or https://",
+    )
+    val imageUrl: String,
```
Hyeon9mak: Request/Response DTOs and Domain Objects are used in the Presentation Layer and Domain Layer, respectively. In your opinion, Deniz, which layer is more appropriate for handling validation?

Hyeon9mak (changes requested):
Hi Deniz!
Hope you're enjoying your weekend!
Thanks to your relaxed timeline for the review, I was also able to enjoy a more laid-back weekend. 😄 

You've done a great job reflecting the requirements for step 2. 👍 
I left a few comments, not about functionality, but regarding separation of responsibilities across layers and some finer details.

As you've probably experienced yourself, the current stage doesn’t really require the complexity of Spring Security.
It might actually be a great chance to get more hands-on with JWT and build a deeper understanding that way.

Also, feel free to take this opportunity to gradually polish the code 😄

src/main/kotlin/ecommerce/controller/MemberController.kt
deniz-oezdemir: Thanks for pointing that out! 🙌
You're totally right, using both "member" and "user" was inconsistent and confusing. I started out with "user" and when "member" was mentioned in the requirements I started using that term too. Now, I've renamed everything from "member(s)" to "user(s)" for clarity and consistency. This includes API paths, class names, and variable names. Let me know if I missed anything! 😊

src/main/kotlin/ecommerce/controller/MemberController.kt
deniz-oezdemir: You’re absolutely right! I moved the default role logic into the `UserService.register()` method so the controller now just passes through the role as-is. It definitely feels cleaner this way. Appreciate the helpful suggestion! 🙌

src/main/kotlin/ecommerce/controller/MemberController.kt
deniz-oezdemir: I moved the login logic to the service layer. Now, `userService.login(email, password)` does the user lookup and password check. The password check is a method on the `User` class. If the credentials are wrong, the service throws a ResponseStatusException with status 403. The controller only handles the request and response. The global exception handler sends the correct HTTP status. Thanks for the suggestion. 🙏

src/main/kotlin/ecommerce/controller/CartController.kt
deniz-oezdemir: Good point! The controller no longer needs the full User entity, especially since fields like password aren't used. I replaced it with LoggedInUser to make the controller depend only on what's necessary for request handling. 😄

src/main/kotlin/ecommerce/config/AdminRoleInterceptor.kt
deniz-oezdemir: UNAUTHORIZED (401) means the user is not authenticated (e.g., no or invalid token). But here, the user is authenticated, we verified the JWT and loaded the user, so returning 401 is misleading.
The more accurate response is FORBIDDEN (403) because the user is authenticated but not permitted to access the resource due to lacking the ADMIN role.
Thanks for pointing it out, I'll update the code to reflect that. 🙏

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
deniz-oezdemir: Yes, `RuntimeException` covers many errors, including server faults. Returning HTTP 400 (Bad Request) here wrongly tells the client they made a mistake. This can confuse users. Instead, I'll return HTTP 500 (Internal Server Error) to clearly show that the problem is on the server side. :-)

src/main/kotlin/ecommerce/config/WebMvcConfig.kt
deniz-oezdemir: Thanks for the suggestion! I agree that naming it `userRoleInterceptor` better reflects its responsibility and is also more consistent with the `adminRoleInterceptor`. 😄

src/main/kotlin/ecommerce/repository/CartItemHistoryRepository.kt
deniz-oezdemir: Thanks, that's a good point. I agree this is a bit raw and inconsistent with how other repositories behave. I've updated the method to return a typed `TopProductStat` DTO now, so it's clearer and more aligned with the rest of the data access layer. 👍

src/main/kotlin/ecommerce/service/ProductService.kt
deniz-oezdemir: Good catch!
validateUniqueName is only used inside ProductService, so there's no reason for it to be public.
I’ve updated it to private. Thanks! 😄

src/main/kotlin/ecommerce/dto/ProductRequest.kt
deniz-oezdemir: Thanks for raising this. I think the presentation layer is the right place for these checks.

Putting validation on the request DTO keeps things simple. It catches bad input before it reaches the service layer. That helps avoid unnecessary logic deeper in the code.

Still, if a rule is part of the business logic, like making sure a price is always above zero in this case, it makes sense to repeat it in the domain layer. That way, we stay safe even if somehow the controller is skipped and calls are directly made to the service directly.

So my take:
- Keep basic format and structure checks in the DTO.
- Add business-critical rules in the domain layer too, even if they overlap.

This helps catch issues if the service is called directly or if input validation changes in the future.

So I have added price validation in `ProductService`'s update and create methods. It may feel a bit redundant but keeps our app more robust overall. What do you think?

deniz-oezdemir:
Hi @Hyeon9mak!

Thanks again for the thoughtful feedback, I really appreciate it. 🙏 

I have implemented all the changes you suggested. I left one conversation unresolved regarding the validations in the Presentation vs. Domain Layer. I would love to hear your thoughts on my approach here.

Let me know what you think!

Thanks, Deniz

src/main/kotlin/ecommerce/controller/UserController.kt
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/api/users")
+class UserController(
+    val userService: UserService,
+    val jwtService: JwtService,
```
Hyeon9mak, line 20: > Now, I've renamed everything from "member(s)" to "user(s)" for clarity and consistency.

We pursue object-oriented programming to improve maintainability.
Consistency, in my view, is also a key factor that contributes to better maintainability. 💯

src/main/kotlin/ecommerce/dto/LoggedInUser.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+data class LoggedInUser(
+    val id: Long,
+    val email: String,
+    val role: String,
+)
```
Hyeon9mak, line 7: >  I replaced it with LoggedInUser to make the controller depend only on what's necessary for request handling. 😄

😄 💯

src/main/kotlin/ecommerce/service/UserService.kt
```diff
@@ -0,0 +1,37 @@
...
+        return userRepository.getByEmail(email)
+    }
+
+    fun login(
+        email: String,
+        providedPassword: String,
+    ): LoggedInUser {
+        val user = getByEmail(email) ?: throw ResponseStatusException(HttpStatus.FORBIDDEN, "Invalid credentials.")
+        if (!user.checkPassword(providedPassword)) throw ResponseStatusException(HttpStatus.FORBIDDEN, "Invalid credentials.")
+        return user.toLoggedInUser()
```
Hyeon9mak: this cases looks like invalid auth info, not permission! 👀

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+    fun handleResponseStatusException(ex: ResponseStatusException): ResponseEntity<Map<String, String>> {
+        val body = mapOf("error" to (ex.reason ?: "Unexpected error"))
+        return ResponseEntity(body, ex.statusCode)
+    }
+
+    @ExceptionHandler(RuntimeException::class)
+    fun handleRuntimeException(ex: RuntimeException): ResponseEntity<Map<String, String>> {
+        val body = mapOf("error" to (ex.message ?: "Unexpected error"))
+        return ResponseEntity(body, HttpStatus.INTERNAL_SERVER_ERROR)
+    }
```
Hyeon9mak: 👍
And, there is a probability of sensitive info in error message.
So It's better not to deliver the message!

src/main/kotlin/ecommerce/dto/ProductRequest.kt
Hyeon9mak: The more validation logic you have, the better—it’s hard to argue against that! 🔥 

But if it goes too far, it can lead to unnecessary validations that might slow down performance (though that’s rare), significantly reduce readability, and increase the number of places you need to modify when business rules change—raising the risk of human error. 😱 

That’s why I believe there should be a balanced compromise.

> Putting validation on the request DTO keeps things simple. It catches bad input before it reaches the service layer. That helps avoid unnecessary logic deeper in the code.

I completely agree with that point as well—early validation makes things much easier to work with.

However, we should also consider it from multiple perspectives.
For example, what if we later add automated jobs like Spring Batch in addition to Admin APIs? 🤔 

Exploring these kinds of questions is, I think, one of the most enjoyable parts of development. 😄

Hyeon9mak: From this point on, this is entirely my personal opinion—please just take it as a reference! 💭 

As the name suggests, DTO (Data Transfer Object) is an object used for transferring data between layers.

For instance, imagine a scenario where we use a single object across Controller, Service, Domain, and Repository layers in an environment where the Controller is frequently changing. Every time the Controller changes, all layers would be affected.
To avoid this kind of ripple effect, we separate layers using DTOs.
The Controller can extract fields from the DTO and pass them to the Service, and the Service can do the same when passing data to the Domain or Repository. This way, changes in the Controller won’t directly impact the other layers.

Because of this, I personally believe DTOs should not contain business logic like “the name must be between 3 and 8 characters.”
They should be lightweight, disposable objects used solely for data transfer between layers, and they shouldn't require separate validation or testing.
If it feels natural to embed business rules into a DTO, then I’d argue it’s no longer a DTO—but rather a signal that you’re looking at a Domain model.

What are your thoughts on this? 😄

src/main/kotlin/ecommerce/service/UserService.kt
```diff
@@ -0,0 +1,37 @@
...
+class UserService(
+    private val userRepository: UserRepository,
+) {
+    fun register(
+        email: String,
+        password: String,
+        role: String?,
+    ): Long {
+        if (userRepository.existsByEmail(email)) throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Email already in use.")
+
```
Hyeon9mak: HttpStatus code is (Obviously) HTTP layer's responsibility. 😉 
What about just throw CustomException?

src/main/kotlin/ecommerce/service/UserService.kt
```diff
@@ -0,0 +1,37 @@
...
+        val user = User(email = email, password = password, role = role ?: "USER")
+        return userRepository.create(user)
+    }
```
Hyeon9mak: I think `User` constructor could have default role parameter.
and it's not a just string. we could declare enum class.

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -0,0 +1,48 @@
...
+    private fun validatePrice(price: Double) {
+        if (price < 0.01) {
+            throw ResponseStatusException(HttpStatus.BAD_REQUEST, "Price must be greater than 0.")
+        }
+    }
```
Hyeon9mak: it's hard to generate test code.
we learned about value class on past missions 😄

Hyeon9mak (changes requested):
Hi Deniz! Sorry for late review 🙏 
Nice work on reflect review comments.

I left some new comments, and answer of ur question!
Lets keep going little more!
Feel free to ask anytime 😄

src/main/kotlin/ecommerce/dto/ProductRequest.kt
deniz-oezdemir: Thanks for sharing your perspective. I can follow your logic and agree that DTOs should remain simple data carriers and not include business logic. Moving validation to the service layer helps keep DTOs lightweight and avoids ripple effects when controllers or services change. This approach also makes the business rules easier to test in one place.

I refactored the code accordingly, removed validation from the DTO and added it to the service layer, along with new tests to cover those validations. By doing this, we keep a clean separation of concerns: DTOs handle data transfer, services enforce business rules, and domain models focus on state and behavior. Overall, this leads to code that’s easier to maintain and evolve.

Thanks for your input! 🙏

src/main/kotlin/ecommerce/service/UserService.kt
deniz-oezdemir: Thanks for the feedback! I updated the login method to throw `401 Unauthorized` instead of `403 Forbidden` for invalid credentials. This makes it clearer that the issue is authentication, not authorization.

src/main/kotlin/ecommerce/config/GlobalExceptionHandler.kt
deniz-oezdemir: Thanks for the feedback! I updated the exception handler to return a generic error message instead of the exception’s message, so we avoid potentially exposing any sensitive information in the response. 👍

src/main/kotlin/ecommerce/service/UserService.kt
deniz-oezdemir: I agree and have moved the HTTP status responsibility out of the service layer by introducing a custom exception `EmailAlreadyUsedException`. This way, the service focuses purely on business logic, and the HTTP status mapping happens centrally in the global exception handler. This keeps the layers clean and easier to maintain.

src/main/kotlin/ecommerce/service/UserService.kt
deniz-oezdemir: Thanks for the suggestion! I replaced the raw role string with a proper enum to enforce valid roles and added a default role parameter in the `User` constructor for convenience. I also updated the repository to save the enum’s string value `role.name` to avoid persistence errors. Additionally, I introduced an `InvalidRoleException` to handle invalid roles gracefully. The controller and tests now use the Role enum directly, improving type safety throughout. 🙏

src/main/kotlin/ecommerce/service/ProductService.kt
deniz-oezdemir: I moved the price validation test directly to a new `Price` value class. Also added a standalone unit test. This keeps validation concerns separated and simplifies the service tests. Thanks for pointing this out 😄

deniz-oezdemir:
Hi @Hyeon9mak,

Thanks for your detailed review! 🙏 

I’ve implemented all the requested changes and resolved the comments accordingly. Apologies for my late response. Instead of working alone on this PR, I wanted to make the most of the time we have together as a team under the week to do pair programming on the next mission.

Thanks for the great reviews! Wishing you a wonderful and restful weekend. 😊

Best, Deniz

src/test/kotlin/ecommerce/entity/PriceTest.kt
```diff
@@ -0,0 +1,15 @@
...
+
+class PriceTest {
+    @Test
+    fun `Price throws if less than 0_01`() {
+        val exception = assertThrows<IllegalArgumentException> {
+            Price(0.0)
+        }
+        assertThat(exception.message).contains("Price must be greater than 0.")
+    }
+}
```
Hyeon9mak, line 15: image

src/test/kotlin/ecommerce/service/ProductServiceTest.kt
```diff
@@ -0,0 +1,96 @@
...
+    @Test
+    fun `createProduct throws if name is longer than 15 characters`() {
+        val longName = "ThisNameIsWayTooLong"
+        val request = ProductRequest(name = longName, price = Price(1.0), imageUrl = "https://valid.url")
+
+        val exception = assertThrows<ResponseStatusException> {
+            productService.createProduct(request)
+        }
+
+        assertThat(exception.reason).contains("Name must be between 1 and 15 characters")
```
Hyeon9mak, line 35: Name also can convert to value class 😉

Hyeon9mak (approved):
Hi Deniz! Great work all the way to the end — you’ve done an excellent job.

I’m glad to see you responded positively to the feedback.
I hope the comments around separating responsibilities across layers were helpful to you. 😆 

Since all the requirements have been met, I’m approving this PR! 👍 
I also hope the feedback I’ve shared so far has been useful.

Wishing you the best as you continue with the remaining missions — enjoy! 😄

src/test/kotlin/ecommerce/service/ProductServiceTest.kt
deniz-oezdemir: Yes, I will leave this comment unresolved to keep it in mind for future missions! Thanks so much for your thorough feedback! 🙏

PR #23, opened 2025-07-25
[spring-ecommerce-product - Step-2] Submit saragrosser's mission

saragrosser:
IMPORTANT NOTE:
I had trouble with my branches so I think all the commits since the first of step1 are here, please note that my implementation for STEP 2 starts on Jul 22, 2025 with the README file for step2-1 thank you!!

Step 2: Validation, Cart System, Authentication, and Admin Analytics
This mission was a big leap forward in applying backend principles in a more realistic and secure context. It allowed me to go beyond CRUD and build a working system with authentication, role-based access control, and statistical endpoints all while learning how to secure and test them properly.

Key Takeaways & Reflections
From public endpoints to secure APIs:
In Step 2.2 and 2.3, I implemented JWT-based authentication for the first time in a Spring Boot application. Adding a custom annotation (@LoginMember) and a corresponding HandlerMethodArgumentResolver taught me how authentication data can be injected cleanly into controller methods. This was a powerful and difficult concept to learn and apply across the cart flow.

Role-based access with interceptors:
In Step 2.4, I introduced an AuthInterceptor to restrict admin-only endpoints. This gave me insight into how Spring can be extended using interceptors for access control, and how to manage route-based restrictions based on user roles.

Testing under real conditions:
Testing is really hard for me, once the authentication was added, all my previous tests broke 😅 but fixing them helped me understand how to register users, log in programmatically, and include tokens in test requests. Writing full integration tests for the cart flow helped me simulate a real-world use case end-to-end, from registration to product interaction.

I ran into several technical issues:
-Mostly my tests were failing and it took me a long time to refactor
-Custom exceptions not being recognized until I registered them in a global handler.
-@LocalServerPort not injecting correctly at first.
-KtLint errors (some auto-fixable, others needing manual renaming).
-Learning how to apply an interceptor
These debugging moments made me more confident with both Spring and the Gradle ecosystem.

I am open to all feedback thank you for reviewing it :)


src/test/kotlin/ecommerce/controller/ProductRestControllerTest.kt
```diff
@@ -0,0 +1,99 @@
...
+                .extract().response()
+
+        println("Status code: ${response.statusCode}")
+        println("Body: ${response.body.asString()}")
+
+        productId = response.jsonPath().getLong("id")
+    }
+
+    @Test
+    fun addProduct() {
```
pkeugine, line 62: This test fails when I run all tests at once.
In our previous step, [there was also a test failure](https://github.com/herotechcourse/spring-ecommerce-product/pull/6#discussion_r2216642366), and I think this is a great opportunity for us to understand the importance of "PR must not contain failing builds or tests".

I hope we get used to this kind of checks before the review process, as it will be much more important in real team projects.

pkeugine (changes requested):
Hello Sara,
There is an issue in this PR, which we previously covered last time.
Please fix this issue, and then request for review again.

src/test/kotlin/ecommerce/controller/ProductRestControllerTest.kt
saragrosser: Hi PK thank you and I apologize when I created the PR all my tests where passing but there was an issue with my build I have corrected it now it should be working and all the tests passing

README.md
```diff
@@ -60,3 +43,116 @@ Database tables must be initialized automatically when the application starts.
...
+
+## Step 2.4 - Cart Statistics & Admin Role
+
+### Features to Implement step 2.4
+
+- [ ] add endpoint to get top 5 most added products in last 30 days
+- [ ] add endpoint to get members who added to cart in last 7 days
+- [ ] restrict `/admin/**` endpoints to users with ADMIN role
+- [ ] return detailed product stats (name, count, last added time)
+- [ ] return member info (id, name, email) for recent cart activity
```
pkeugine: Nice work on updating the README 👍 
It would be great for you to check the todo list in step 2.4 and see if we can mark any of them as finished. :)

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,103 @@
...
+                    errors = listOf(FieldError("name", "Product name must be unique")),
+                )
+            return ResponseEntity.badRequest().body(error)
+        }
+
+        val product = request.toEntity()
+        productRepository.createProduct(product)
+
+        // val saved = productRepository.getAll().last()
+        return ResponseEntity.ok().body(product)
```
pkeugine: I checked the commit that was added in order to make the test pass.
It would be better if we could fix the issue by removing the logic that caused it, instead of commenting it out.
We won't be needing it anyways, so we might as well not let it get in the main branch.
Please also check other parts where there are commented logic.

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,103 @@
...
+            val error =
+                ErrorResponse(
+                    message = "Validation failed",
+                    errors =
+                        listOf(
+                            FieldError("name", "Product name must be unique"),
+                        ),
+                )
+            return ResponseEntity.badRequest().body(error)
+        }
```
pkeugine: Since we have `GlobalExceptionHandler`, why don't we let it handle this `badRequest` with `ErrorResponse`?
It would help us maintain failed reponses in one place!
Please check on other places where we return `badRequest()` in controller.

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        val updatedProduct =
+            existingProduct.copy(
+                name = request.name,
+                price = request.price,
+                imageUrl = request.imageUrl,
+            )
+
+        productRepository.updateProduct(updatedProduct)
```
pkeugine: Also, I'm curious about `when you create a service class`.
This is because I noticed you've created `CartService` and `MemberService`, but there isn't a `ProductService`.
I'm just asking to know your thoughts behind this, so don't feel too pressured by this question :)

src/main/kotlin/ecommerce/controller/ProductRestController.kt
```diff
@@ -0,0 +1,103 @@
...
+
+        // val saved = productRepository.getAll().last()
+        return ResponseEntity.ok().body(product)
+    }
+
+    @PutMapping("/{id}")
+    fun update(
+        @PathVariable id: Long,
+        @RequestBody @Valid request: ProductRequest,
+    ): ResponseEntity<Any> {
```
pkeugine: By moving exception related responses to `GlobalExceptionHandler`, we'll also be able to change the type in `ResponseEntity` as well 👍.

This is just a tip since we're going over generics : 
When you have the time, I strongly recommend going over how generics work in Kotlin.
It will be used a lot as we develop features using Kotlin, and it is also a 'must' in interview questions. :)

src/main/kotlin/ecommerce/dto/TokenRequest.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.dto
+
+data class TokenRequest(
+    val email: String,
+    val password: String,
+)
```
pkeugine: Since this class is not being used, let's delete it!
Also, I think it would be great for us to unify the name to `Member` or `Token` for `TokenResponse`,
since it's getting used with `MemberRequest`. So it's either both `TokenRequest` and `TokenResponse`, or `MemberRequest` and `MemberResponse`.

src/main/kotlin/ecommerce/mapper/ProductMapper.kt
```diff
@@ -0,0 +1,13 @@
...
+import ecommerce.model.Product
+
+fun ProductRequest.toEntity(): Product {
+    return Product(
+        id = 0,
+        name = this.name,
+        price = this.price,
+        imageUrl = this.imageUrl,
+    )
+}
```
pkeugine: I'm curious about this decision where `toEntity()` is in `ProductMapper` instead of `ProductRequest` itself. :)

src/main/kotlin/ecommerce/mapper/ProductMapper.kt
```diff
@@ -0,0 +1,13 @@
+package ecommerce.mapper
+
+import ecommerce.dto.ProductRequest
+import ecommerce.model.Product
+
+fun ProductRequest.toEntity(): Product {
+    return Product(
+        id = 0,
```
pkeugine: Here is a fun (?) little question! :  

`Product`'s id field is currently `Long` instead of `Long?`. This means that id field cannot be null even when there is no value of it. As a result, we have to give it a value, so in our case, it's 0. Do you think this approach is better than utilizing `null`? Why or why not?

src/main/kotlin/ecommerce/controller/CartAdminController.kt
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @GetMapping("/active-members")
+    fun getActiveMembers(
+        @LoginMember member: Member,
+    ): ResponseEntity<Any> {
+        val result = cartService.getRecentlyActiveMembers()
+        return ResponseEntity.ok(result)
+    }
+}
```
pkeugine: Generic types in these `ResponseEntity` could be changed from `Any` to respectable dto classes, just like what we covered in the previous comment :)

src/main/kotlin/ecommerce/dto/ProductStatResponse.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.dto
+
+data class ProductStatResponse(
+    val name: String,
+    val count: Int,
+    val lastAddedAt: String,
+)
```
pkeugine, line 7: Currently, dto classes are used in repositories, services, and controllers.

`Request` and `Response` dto classes are pretty interesting, because they not only carry the data around layers, but also validate inputs and are tightly coupled with Spring Boot MVC. If we were to change the controller layer to console friendly `OutputView` (remember this guy? 😄 ), we might have to update our `Request` and `Response` classes, which would also affect repositories and services.

This is my personal opinion : I think it would be great if we only used `Request` and `Response` dto classes in controllers, and let services & repositories handle data transfer in their own way (by returning entities, or by using a new dto, etc).

How do you think about this? There are developers who think different, so feel free to share your personal thoughts. :)

src/main/kotlin/ecommerce/model/CartItem.kt
```diff
@@ -0,0 +1,10 @@
+package ecommerce.model
+
+import java.time.LocalDateTime
+
+data class CartItem(
+    val id: Long,
+    val memberId: Long,
+    val productId: Long,
+    val createdAt: LocalDateTime,
```
pkeugine, line 9: This is my personal preference when it comes to naming:
I like to make things simple and constant. This means when we're referring to a `product` in a `cart`,
I would call it `CartProuduct` instead of `CartItem`.
This is because when project becomes bigger and as more and more entities get created, there will be so many names for us to take care of, and I tend to forget that `Item`, `Product`, `itemId`, `productId` etc... are the same thing. By unifying the name, we get to remember more easily, and we can get help of searching without overlooking.

But as I said, this is just my preference, so feel free to keep it as you like 😄

src/main/kotlin/ecommerce/service/MemberService.kt
```diff
@@ -0,0 +1,46 @@
...
+
+        val saved = memberRepository.save(Member(email = request.email, password = request.password))
+        val token = jwtTokenProvider.createToken(saved.id)
+        return TokenResponse(token)
+    }
+
+    fun login(request: MemberRequest): TokenResponse {
+        val member =
+            memberRepository.findByEmail(request.email)
+                ?: throw IllegalArgumentException("Invalid email or password")
```
pkeugine: This comment is just out of curiosity :  

`lines 16 ~ 19` and `lines 27 ~ 29` do almost the same thing (but acts different).
But their line break and logic process seem different!
Do you have any certain convention that you follow, or could it be formatter that changed it? 🤔

src/main/kotlin/ecommerce/service/MemberService.kt
```diff
@@ -0,0 +1,46 @@
...
+
+        if (member.password != request.password) {
+            throw IllegalArgumentException("Invalid email or password")
+        }
```
pkeugine, line 33: Nice work on not making it clear rather it's email that's wrong, or password that's incorrect 💯 
When it comes to security, being abstract and not nice tends to be a better call. I would've done the same thing 😄 .

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
```diff
@@ -0,0 +1,38 @@
...
+        response: HttpServletResponse,
+        handler: Any,
+    ): Boolean {
+        val token = request.getHeader("Authorization")?.removePrefix("Bearer ") ?: return unauthorized(response)
+
+        val member =
+            try {
+                memberService.findByToken(token)
+            } catch (e: Exception) {
+                return unauthorized(response)
```
pkeugine, line 24: This is an interesting, and also very important question :  
What's the difference between authentication and authorization?

src/test/kotlin/ecommerce/CartIntegrationTest.kt
```diff
@@ -0,0 +1,158 @@
...
+import org.springframework.boot.test.web.server.LocalServerPort
+import org.springframework.http.HttpStatus
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.test.annotation.DirtiesContext
+import org.springframework.test.context.ActiveProfiles
+
+@ActiveProfiles("test")
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
pkeugine: It's interesting to see all these annotations in test! We've seen `@SpringBootTest` and `@DirtiesContext` previously, but other two seems new. Could you tell me what each of these do?

Also, I think we'll be able to remove some if we tweak the test a little. By removing, we'll be able to enhance the test results (such as speed). Why don't we give it a try?

src/test/kotlin/ecommerce/controller/ProductRestControllerTest.kt
```diff
@@ -0,0 +1,99 @@
...
+            RestAssured
+                .given().log().all()
+                .contentType(ContentType.JSON)
+                .body(productJson)
+                .`when`().post("/api/products")
+                .then().log().all()
+                .extract().response()
+
+        println("Status code: ${response.statusCode}")
+        println("Body: ${response.body.asString()}")
```
pkeugine: Why don't we remove unnecessary `println` in tests?
Since we decided to `log().all()` values that go in and out of `RestAssured`,
we'll be able to read them in logs even if we don't print them out.

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
```diff
@@ -54,33 +67,46 @@ class ProductRepositoryTest {
...
-            ('Notebook', 3.49, 'https://upload.wikimedia.org/wikipedia/commons/6/6e/Moleskine_notebook.jpg');
+            INSERT INTO products (name, price, image_url) VALUES ('Coffee', 10.99, 'https://upload.wikimedia.org/wikipedia/commons/thumb/c/c5/Roasted_coffee_beans.jpg/1200px-Roasted_coffee_beans.jpg');
+            INSERT INTO products (name, price, image_url) VALUES ('Hand cream', 6.99, 'https://cdn.idealo.com/folder/Product/6178/3/6178302/s1_produktbild_gross/neutrogena-hydro-boost-creme-gel-50ml.jpg');
+            INSERT INTO products (name, price, image_url) VALUES ('Chair', 77.99, 'https://encrypted-tbn1.gstatic.com/shopping?q=tbn:ANd9GcRqUHblIQ5QCehPdDXFHqSqUKqxNExCjw_mKe-wsYJYNXP7S7pnlrGpBKEPGzutl3CSIxRQSJjb3HbbQevqNeDMZQdxtP0ml591k0zsn-b-4KAMs9aMwJngvHNPGf7KHbNhUuOn6g&usqp=CAc');
+            INSERT INTO products (name, price, image_url) VALUES ('Notebook', 3.49, 'https://upload.wikimedia.org/wikipedia/commons/6/6e/Moleskine_notebook.jpg');
```
pkeugine, line 121: What other ways could we create data for each test? It would be great for us to look it up, and compare them to find the best one that fits our situation :)

README.md
```diff
@@ -60,3 +43,116 @@ Database tables must be initialized automatically when the application starts.
...
+
+
```
pkeugine, line 158: **PR message : Testing is really hard for me, once the authentication was added, all my previous tests broke**

😄 That happens quite often for me too. As we add more features into the project, tests also become things that we need to manage. In other words, tests become `legacy code` just like application logic. If we don't manage them properly, they'll just end up being disabled, or be a mess. A mess SO BIG that it's so hard to refactor.

Maybe that's why we're trying to learn how to write good tests, try to manage them as we make logic, and approach in a different way such as using TDD.

That makes me curious, what do you think are the most important kinds of tests? Unit test, integration test, or even E2E test? What IS an E2E test anyway? 🤔

pkeugine (changes requested):
Hello Sara, and nice work on step 2.
The project was very easy to follow 👍 

I left some reviews, questions, and interesting topics. It would be great if we could talk about it, and would be even better if you could share it with your fellow crews and ask what they think about it! Most of the questions don't really have a one perfect answer, so feel free to build your own opinion around it.

If you're finished with reviews, go ahead and request another review! I'll be waiting for it. Leave a comment or DM me any time if you have any questions.

Happy Hacking! 👍 👍 👍

README.md
saragrosser: Thank you this are the type of newbie mistakes I still make like forgetting to update README, thank you for pointing it out!

src/main/kotlin/ecommerce/controller/ProductRestController.kt
saragrosser: Thank you PK, I left this commented out because I was unsure if it made sense and that way you could see what I was doing before, I have deleted it now since it is not needed anyway :)

src/main/kotlin/ecommerce/controller/ProductRestController.kt
saragrosser: It is so funny how I still manage to mix responsibilities haha, but I love how every time my code looks better and better, I have refactor this and now my controller focuses only on business logic, and all error responses are unified in GlobalExceptionHandler! :) thank you !!!

src/main/kotlin/ecommerce/controller/ProductRestController.kt
saragrosser: The honest answer is that when we started implementing step1 we did not had in mind the service layer at all, and then when step 2 started I needed it for member and for cart but with all the things I was working on I forgot about Product Service layer, there is no special reason behind it. 

That is why I decided to now implement it to also follow more coherence in the code and readability, this will allow me to re-use product logic elsewhere if needed in the future.

src/main/kotlin/ecommerce/controller/ProductRestController.kt
saragrosser: Initially I used ResponseEntity<Any> because I was manually returning either a Product or an ErrorResponse inside the controller.

But now that I refactor it all validation and error handling is moved to GlobalExceptionHandler, the controller methods have a consistent return type, so it makes much more sense to specify the exact generic type (like ResponseEntity<Product>).

After reading a bit I learned that not declaring a specific type creates ambiguity and that hurts maintainability and clarity, basically you don't know what to expect. I appreciate the tip about Kotlin generics! I’ll definitely take the time to study them more deeply especially how they relate to type safety and API contracts. It’s already making more sense now!

src/main/kotlin/ecommerce/dto/TokenRequest.kt
saragrosser: Thank you PK I have deleted the unused class

For the naming part I created a hybrid = "TokenMemberRequest"  because for me it makes sense to mention both so this way I create coherence but also for my understanding I keep both concepts, I hope my logic is ok

src/main/kotlin/ecommerce/mapper/ProductMapper.kt
saragrosser: Great question and to be honest, I wasn’t 100% sure at the time but because I am trying to learn and keep the SRP, I think I instinctively wanted to keep DTOs clean and avoid mixing them with mapping logic.

Since mapping might evolve later, I thought separating it into ProductMapper.kt would give me more flexibility and keep the code cleaner.

But I’d love to hear your thoughts especially if you think the mapping would be better located in the DTO itself in this context!

src/main/kotlin/ecommerce/mapper/ProductMapper.kt
saragrosser: That’s such a great point, I initially used id = 0 because I was working with  simple manual JDBC and wanted to avoid null-handling overhead.

But now that I’m transitioning the project toward a more ORM-based setup with JPA, it actually makes more sense to refactor id to Long? to better reflect the semantics of a not-yet-persisted entity.

src/main/kotlin/ecommerce/controller/CartAdminController.kt
saragrosser: Thank you I have refactor the CartAdminController they now have a defined type List<CartItem>

I noticed also my CartController was having an undefined ResponseEntity type so I also refactor that!

src/main/kotlin/ecommerce/dto/ProductStatResponse.kt
saragrosser: I hadn’t thought much about the consequences of letting response DTOs flow all the way into the service and repository layers, but your point makes a lot of sense. I can see how it could introduce tight coupling to the controller layer’s format, and how that could create issues if the presentation changes

Im not sure how but I’d be open to refactoring this by maybe introducing internal domain-specific DTOs or just returning domain models (like ProductStat) from the repo/service layers, and only mapping to ProductStatResponse at the controller level?

src/main/kotlin/ecommerce/model/CartItem.kt
saragrosser: Thank you PK, naming consistency is so important, especially as the project grows and more entities are added.

In this case though, I felt like CartItem fits a bit better semantically, since it's more than just a Product, it also carries the memberId, createdAt, and kind of represents a "line item" in the cart rather than the product itself, it makes more sense for me and actually it helps me remember what is inside and the difference. 

So I’ll like to stick with CartItem for now if that is okay, but I’ll definitely keep the naming convention in mind!

src/main/kotlin/ecommerce/service/MemberService.kt
saragrosser: Ahh I see what you mean now thanks for pointing that out! I didn’t really follow a strict convention here I think it just happened naturally or was influenced by auto-formatting.

But I’ll align the formatting to keep it more consistent going forward in the project thank you!

src/main/kotlin/ecommerce/service/MemberService.kt
saragrosser: Thanks! That was intentional I remembered the recommendation to avoid giving attackers clues about valid accounts. Glad to hear we're aligned on that!

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
saragrosser: As I understand it: 

Authentication -> is verifying who you are (if you are who you say you are)
Authorization -> is determining what you’re allowed to do (the permission you have)

src/test/kotlin/ecommerce/CartIntegrationTest.kt
saragrosser: Hi PK to be honest Testing is the hardest thing for me I have really a hard time making my tests pass so the honest answer is I tried all annotations until the test worked because I tried everything 

Here is why: 

@ActiveProfiles("test") -> Ensures Spring loads application-test.yml
@SpringBootTest(...) ->	Spins up entire Spring context with real server
@DirtiesContext(...) -> Re-creates Spring context before every test
@TestInstance(PER_CLASS)-> Allows sharing `BeforeAll` state across test methods

 I had added those extra annotations when I was troubleshooting weird test state issues, but now that things are stable and @BeforeEach handles DB resets, I agree they’re redundant.

I have removed @DirtiesContext and @TestInstance(PER_CLASS), and things are still passing (and faster!). Thanks for the push to simplify!

src/test/kotlin/ecommerce/controller/ProductRestControllerTest.kt
saragrosser: You’re right, those were leftovers from when the test was failing and I needed visibility. Now that we use log().all(), they’re totally redundant.  I have removed them now

src/test/kotlin/ecommerce/repository/ProductRepositoryTest.kt
saragrosser: Thank you for this feedback PK,  I’ve refactored the test data setup to avoid raw SQL strings and replaced them with a helper function that inserts test products using JdbcTemplate.update. This makes the test cleaner and more maintainable while keeping the repository logic isolated from the test setup.

README.md
saragrosser: Yes I have been studying a lot about tests , I think the answer is it depends what are you trying to test to decide which is the most important test, to make a good test you have to answer: what do you really want to verify in this test? (From known to unknown) 

I hope I could show you my notes, but I have a summary of all test types but I will summarize here for you what I have studied: 

1. E2E means End to End Tests -> verify the full request flow from controller to database, it simulates real HTTP requests through the whole process 

2.Slice Test -> test a specific layer in isolation, other layers are replace with mock objects to improve focus you can have 3 types : DAO layer testing (data access layer Respository + DB) Web layer testing (Controller + HTTP responses) Service layer testing (Buisiness logic in Service layer) 

3.Integration Test -> verify real interaction between service and Respository (uses real database)

src/main/kotlin/ecommerce/controller/ProductRestController.kt
pkeugine: I'm sorry if I sound kinda strict when it comes to broken PR or adjustments,
I just wanted to deliver the importance(?) of preventing it :)

Also, thanks for thinking about it, and I'll tell you my secret (?)
since we have git histories and all the logs of previous works,
I can see what you were doing before, what the problem was (by checking out and running tests in previous snapshot), and track down all the process that led to this final state 😂 

Pretty useful in actually work, to be honest.
I definitely recommend studying deeply about git once you have the time.

src/main/kotlin/ecommerce/controller/ProductRestController.kt
pkeugine: I can also see that you're getting better and better, with speed 🚀

src/main/kotlin/ecommerce/controller/ProductRestController.kt
pkeugine: Perfect! 💯

src/main/kotlin/ecommerce/dto/TokenRequest.kt
pkeugine: I'm fine with it 👍

src/main/kotlin/ecommerce/mapper/ProductMapper.kt
pkeugine: It's very fun to read your opinion. I feel the same when making these decisions. :)

In our case, since we don’t have many DTO classes or complex mapping logic yet, I think putting `toEntity()` inside the DTO is the simpler choice with fewer classes to manage and less cognitive overhead.

If we reach a point where the mapping becomes complex or the DTOs/entities start to diverge in structure, I’d definitely consider moving to a dedicated Mapper (especially to keep SRP and testability clean). But for now, having the transformation logic in the DTO makes the intent easy to discover and reduces boilerplate in my opinion. 🤔

src/main/kotlin/ecommerce/dto/ProductStatResponse.kt
pkeugine: I recommend trying out your own way and experience the outcome. I personally think that is the best way to absorb knowledge on this topic, since like I said, there are developers who think different from me.

In current mission, you don't have to worry too much about refactoring this, but it would definitely be a plus if we keep this topic in our head as we maintain multiple projects throughout this course :)

src/main/kotlin/ecommerce/model/CartItem.kt
pkeugine: Sure!
I suggest, and you decide 👍

src/main/kotlin/ecommerce/service/MemberService.kt
pkeugine: It just caught my eyes 😂 
It happens to me too.

src/main/kotlin/ecommerce/auth/AuthInterceptor.kt
pkeugine: Very clean and on point! 💯

src/test/kotlin/ecommerce/CartIntegrationTest.kt
pkeugine: That's the process of learning :)
Let's not forget about [this comment](https://github.com/herotechcourse/spring-ecommerce-product/pull/6#discussion_r2225621368) though!

README.md
pkeugine: I can see you're in a road of rapid learning + growing.
Let's keep it up! 💪

pkeugine (approved):
Hello Sari, 
it was nice to read your honest opinion, thoughts, and see you get better in each of the comments 👍 

Most of the comments I've left are reactions or answers to your questions,
so I think this would be a great time for us to finish this mission and move on.

I can see you're gaining knowledge very fast, and I hope you continue on with it through the next mission!

Thanks for your hard work, and as always,
Happy Hacking! 🚀 🚀 🚀

PR #24, opened 2025-07-26
[spring-ecommerce-product - Step-2] Submit Vito's mission

VittorioDeMarzi:
Hi Dani :) 👋

First of all, huge thanks for all your ongoing support!
This mission has been intense but incredibly rewarding, I’m learning so much along the way.

This PR includes the full backend for a basic eCommerce app, featuring product management, user authentication, cart functionality, and admin-level analytics.

### Product CRUD + Validation (Admin Only)
  Full support for creating, updating, retrieving, and deleting products.
  - GET /api/products — Public product list
  - POST /api/products — Create product (admin only)
  - PUT /api/products/{id} — Update product (admin only)
  - DELETE /api/products/{id} — Delete product (admin only)
  
Validation Rules:
- Name ≤ 15 characters
- Allowed characters: alphanumeric + ()[]+-&/_
- Must be unique
- Price must be > 0
- imageUrl must start with http:// or https://

 Invalid input returns structured, readable error messages.

### Member Registration & Login with JWT
- POST /api/register — Register a new user and receive JWT
- POST /api/login — Authenticate and receive JWT

 Stateless authentication with Authorization: Bearer <token> header.
A custom interceptor checks and validates the token on protected routes.

### Cart Management for Authenticated Users
- GET /api/wishes — View current user’s cart
- POST /api/wishes — Add a product to cart (productId, quantity)
- PUT /api/wishes/{productId} — Update quantity (0 = remove)
- DELETE /api/wishes/{productId} — Remove product from cart

 A cart is automatically created for the user on first use.
All operations require a valid JWT token.

### Admin-Only Statistics
GET /api/admin/stats/top-products
→ Top 5 most-added products in the last 30 days
→ Sorted by number of additions, then by most recent timestamp
→ Response includes: product name, count, and last-added time

GET /api/admin/stats/active-members
→ Members who added items to cart in the last 7 days
→ Response includes: member ID, email

 Restricted to admin users only via custom annotation
 Security & Access Control

### Custom Security Infrastructure
- @LoginMember
Injects the authenticated Member directly into controller methods
Backed by a LoginMemberArgumentResolver

- @CheckAdminOnly
Marks controllers as admin-only
Enforced via AdminCheckResolver

- JwtAuthInterceptor
Intercepts every request and validates JWT token

- LoginMemberArgumentResolver
Resolves the @LoginMember annotation
Fetches the member from the database using the injected email
Simplifies controller logic and avoids repeated lookups

- AdminCheckArgumentResolver
Resolves the @CheckAdminOnly annotation
Verifies the user’s role is ADMIN before proceeding

- WebMvcConfigurer
Registers the JWT interceptor
Adds both argument resolvers (LoginMember, AdminCheck)
Applies interceptor only on selected paths

### Test Status
Some test classes are still missing, especially for the new admin statistics.
However, I’m sending this PR now so I can gather your feedback before Monday and refine what’s missing with more clarity.

Thanks again for your time and for the amazing support 🙏
As always, I’m open to any suggestions, improvements, or ideas you may have!

Looking forward to your feedback!


README.md
```diff
@@ -25,47 +30,237 @@ Content-Type: application/json
...
-  - Handle ProductNotFoundException with a 404 Not Found response.
-  - Handle ProductCreationException and ProductUpdateException with 500 Internal Server Error.
-  - Catch generic RuntimeException and return a 400 Bad Request.
-- [x]Return consistent error responses using a custom ErrorMessageModel
+    - Handle ProductNotFoundException with a 404 Not Found response.
+    - Handle ProductCreationException and ProductUpdateException with 500 Internal Server Error.
+    - Catch generic RuntimeException and return a 400 Bad Request.
+- [x] Return consistent error responses using a custom ErrorMessageModel
+
+# Step 2-1
```
da-nyee, line 84: Nice update 👍

src/main/kotlin/ecommerce/controller/AdminProductController.kt
```diff
@@ -16,35 +18,39 @@ import org.springframework.web.bind.annotation.RequestMapping
...
 
     @PutMapping("/{id}")
     fun updateProduct(
         @PathVariable id: Long,
-        @RequestBody product: ProductRequest,
+        @Valid @RequestBody product: ProductRequest,
+        @AdminOnly member: MemberDto,
     ): ResponseEntity<Void> {
         productService.updateProduct(id, product)
         return ResponseEntity.noContent().build()
```
da-nyee: What about using 200 instead of 204?

src/main/kotlin/ecommerce/dto/ProductRequest.kt
```diff
@@ -1,3 +1,24 @@
...
+import jakarta.validation.constraints.NotBlank
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Positive
+import jakarta.validation.constraints.Size
+
+data class ProductRequest(
+    @field:NotBlank(message = "Name must not be blank")
+    @field:Size(max = 15, message = "Name must be at most 15 characters long")
+    @field:Pattern(
+        regexp = "^[\\w\\s\\(\\)\\[\\]\\+\\-\\&/_]*\$",
```
da-nyee:
```suggestion
        regexp = "^[\\w\\s()\\[\\]+\\-&/_]*$",
```

It seems like there are some redundant escapes 👀

src/main/kotlin/ecommerce/service/ProductService.kt
```diff
@@ -2,20 +2,57 @@ package ecommerce.service
...
+    }
 
-    fun createProduct(productRequest: ProductRequest): Boolean
+    fun createProduct(productRequest: ProductRequest): Boolean {
+        if (productRepository.existsByName(
+                productRequest.name,
+            )
+        ) {
+            throw ProductAlreadyInDBException("Product already exists with name: ${productRequest.name}")
+        }
```
da-nyee: image

You followed the rule 👍

da-nyee:
```suggestion
        if (productRepository.existsByName(productRequest.name)) {
            throw ProductAlreadyInDBException("Product already exists with name: ${productRequest.name}")
        }
```

Just a small suggestion — how about removing some newlines to improve readability?

src/main/kotlin/ecommerce/dto/ProductRequest.kt
```diff
@@ -1,3 +1,24 @@
...
+    )
+    val name: String,
+    @field:Positive(message = "Price must be greater than 0")
+    val price: Double,
+    @field:NotBlank(message = "Image URL must not be blank")
+    @field:Pattern(
+        regexp = "^(https?|ftp)://[\\w.-]+(?:\\.[\\w.-]+)+[/#?]?.*\$",
+        message = "Invalid URL format, example: https://example.com/image.jpg",
+    )
+    val imageUrl: String,
```
da-nyee, line 23: 👍

src/main/kotlin/ecommerce/annotation/LoginMember.kt
```diff
@@ -0,0 +1,5 @@
+package ecommerce.annotation
+
+@Target(AnnotationTarget.VALUE_PARAMETER)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class LoginMember
```
da-nyee, line 5: Custom annotation 👍

src/main/kotlin/ecommerce/configuration/JwtAuthInterceptor.kt
```diff
@@ -0,0 +1,26 @@
...
+class JwtAuthInterceptor(
+    private val jwtTokenProvider: JwtTokenProvider,
+) : HandlerInterceptor {
+    override fun preHandle(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        handler: Any,
+    ): Boolean {
+        val accessToken = request.getHeader("Authorization")?.substringAfter("Bearer ")
+        if (accessToken == null) throw UnauthorizedException("Invalid Token")
```
da-nyee:
```suggestion
        val accessToken = request.getHeader("Authorization")?.substringAfter("Bearer ")
            ?: throw UnauthorizedException("Invalid Token")
```

It could be more idiomatic Kotlin 👀

src/main/kotlin/ecommerce/configuration/JwtAuthInterceptor.kt
```diff
@@ -0,0 +1,26 @@
+package ecommerce.configuration
+
+import ecommerce.exception.UnauthorizedException
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.stereotype.Component
+import org.springframework.web.servlet.HandlerInterceptor
+
+@Component
+class JwtAuthInterceptor(
```
da-nyee, line 10: It would be helpful to study how `DispatcherServlet` works ☺️

src/main/kotlin/ecommerce/configuration/AdminOnlyResolver.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.springframework.stereotype.Component
+import org.springframework.web.bind.support.WebDataBinderFactory
+import org.springframework.web.context.request.NativeWebRequest
+import org.springframework.web.method.support.HandlerMethodArgumentResolver
+import org.springframework.web.method.support.ModelAndViewContainer
+
+@Component
+class AdminOnlyResolver(
+    private val memberRepository: MemberRepository,
+) : HandlerMethodArgumentResolver {
```
da-nyee, line 18: Good use of `HandlerMethodArgumentResolver` and `HandlerInterceptor` 👍

da-nyee: I have one question — what's the difference between `Authentication` and `Authorization`?
I'd like to hear your take on it 👀

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -46,4 +47,83 @@ class GlobalExceptionHandler {
...
+
+    @ExceptionHandler(MemberEmailAlreadyExistsException::class)
+    fun handleEmailAlreadyExistsException(ex: MemberEmailAlreadyExistsException): ResponseEntity<ErrorMessageModel> {
+        val errorMessage =
+            ErrorMessageModel(
+                HttpStatus.CONFLICT.value(),
+                ex.message,
+            )
+        return ResponseEntity(errorMessage, HttpStatus.CONFLICT)
+    }
```
da-nyee:
```suggestion
    @ExceptionHandler(value = [
        ProductAlreadyInDBException::class,
        MemberEmailAlreadyExistsException::class,
    ])
    fun handleConflictException(ex: RuntimeException): ResponseEntity<ErrorMessageModel> {
        val errorMessage =
            ErrorMessageModel(
                HttpStatus.CONFLICT.value(),
                ex.message,
            )
        return ResponseEntity(errorMessage, HttpStatus.CONFLICT)
    }
```

We can handle multiple exceptions like this using a single method 👀

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+            SELECT 
+                p.name AS product_name,
+                COUNT(*) AS added_count,
+                MAX(ci.created_at) AS last_added_at
+            FROM cart_history ci
+            JOIN products p ON ci.product_id = p.id
+            WHERE ci.created_at >= CURRENT_DATE - INTERVAL '30' DAY
+            GROUP BY ci.product_id, p.name
+            ORDER BY added_count DESC, last_added_at DESC
+            LIMIT 5
```
da-nyee, line 45: 👍

da-nyee: If you were to add indexes to `cart_history` and `products` for better performance, how would you do it?

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+                "status" to "ADDED",
+                "created_at" to now,
+            )
+        return insert.executeAndReturnKey(productMap).toLong()
+    }
+
+    fun getTop5ActiveUsers(): List<ActiveUsersResponse> {
+        val sql =
+            """
+            SELECT DISTINCT m.ID, m.EMAIL FROM MEMBERS m
```
da-nyee: image

I think you missed one thing 👀

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+            JOIN CARTS c ON m.id = c.MEMBER_ID
+            RIGHT JOIN CART_HISTORY ch ON ch.CART_ID = c.ID
```
da-nyee: Is it okay to use `RIGHT JOIN`?

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
```diff
@@ -0,0 +1,82 @@
...
+            WHERE ch.CREATED_AT >= CURRENT_DATE - INTERVAL '7' DAY
+            LIMIT 5
```
da-nyee: I think it's better to use `ORDER BY` as well, since it might not be sorting — or working — as expected 👀

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
```diff
@@ -0,0 +1,157 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+class AdminControllerTest {
```
da-nyee, line 20: > Some test classes are still missing, especially for the new admin statistics.
However, I’m sending this PR now so I can gather your feedback before Monday and refine what’s missing with more clarity.

Looking forward to the additional tests you'll be adding soon 😄

da-nyee: By the way, have you had any difficulties with writing E2E tests?

da-nyee (changes requested):
Hi Vito! 🙌
Good job on this step — there were quite a lot of requirements, but you handled them well 💯

I went through it and left a few comments — feel free to check them out.
If anything’s unclear or if there’s anything else you’d like me to take a closer look at, just let me know 🙇‍♀️

src/main/kotlin/ecommerce/controller/AdminProductController.kt
VittorioDeMarzi: You're absolutely right to bring it up. From what I understand, 204 (No Content) is typically used when the server successfully processes the request but doesn’t return any content in the response body.
Since I’m currently not returning a ProductDto, I thought 204 would be more appropriate in this case.

That said, I’m happy to switch to 200 if you think it makes more sense, curious to hear your perspective on this!

src/main/kotlin/ecommerce/service/ProductService.kt
VittorioDeMarzi: Much more clear 😄

src/main/kotlin/ecommerce/configuration/JwtAuthInterceptor.kt
VittorioDeMarzi: I will learn to pay attention to this small details. Thanks for pointing this out :)

src/main/kotlin/ecommerce/configuration/AdminOnlyResolver.kt
VittorioDeMarzi: `Authentication` determines whether users are who they claim to be.
If the provided credentials are correct, the system assumes the identity is valid and grants access, like in this case, where a valid login returns a token.

`Authorization` determines what users can and cannot access.
It's usually done after successful authentication. In this project, it determines what actions an Admin or a regular User can perform:
- Admin is able to create, update, delete a product
- User is able to have access to the cart and add/remove/update products to the cart

VittorioDeMarzi: By the way, I have a question for you:

As you can se the `LoginMemberArgumentResolver` and the `AdminOnlyResolver` are quite similar and there is code duplication since the only difference is that `AdminOnlyResolver` che the role after finding the member in the db. 

Do you think there's a clean way to avoid this duplication? Maybe some shared base logic or abstraction?

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
VittorioDeMarzi: This was a great tip 😃
I refactored the handler as suggested to group similar exceptions and reduce duplication, definitely cleaner now. Thanks!

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
VittorioDeMarzi: I did't want to refactor everything but here it is 😜 Hope it looks better now 😊

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
VittorioDeMarzi: Good catch, I was experimenting with different JOIN types and accidentally committed the wrong one. You're right, a regular INNER JOIN makes more sense here.

In this case, we're only interested in members who actually added items to their cart in the last 7 days. Using an INNER JOIN ensures we only get records where there’s a match between MEMBERS, CARTS, and CART_HISTORY 😁

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
VittorioDeMarzi: You're absolutely right, without ORDER BY, the results can be unpredictable.
In fact, the task didn’t require limiting to 5 users, so I removed the LIMIT clause and added ORDER BY m.ID to keep the results consistent. Thanks for the reminder 👀

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
VittorioDeMarzi: I’d say the trickiest part was setting up the @BeforeEach properly, especially cleaning up all the tables except for members and making sure I could log in with the admin and get the token back, to send it with the request.
Once authentication was working, writing the actual tests for each endpoint came pretty naturally 😄

That said, the tests still rely on the admin already existing in the database, so if the data changes, they'll break.

I also introduced AdminStatsControllerTest. In this case, I'm relying on specific dates when products were added to the cart, so the test will likely fail in a couple of weeks.
Do you have any suggestions on how to make this more robust and avoid failures over time?

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
VittorioDeMarzi: Could I index with this query in my schema?

CREATE INDEX idx_cart_history_product_created ON cart_history(product_id, created_at);
Since I am grouping by product_id and ordering by created_at

I really appreciate your feedback on this :)

VittorioDeMarzi:
Hi @da-nyee! 🌟

Thanks again for the thorough review and the thoughtful suggestions.
I’ve addressed all the comments:
- Switched to INNER JOIN and cleaned up the query logic ✅
- Applied your suggestions on simplifying conditionals and regex escapes ✅
- Refactored exception handlers to avoid duplication ✅
- Adjusted the LIMIT/ORDER BY logic and removed unnecessary clauses ✅
- Added the missing test class for admin stats ✅

Let me know if there's anything else you'd like me to fix or improve

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -27,8 +28,14 @@ class GlobalExceptionHandler {
...
-    @ExceptionHandler(ProductCreationException::class)
-    fun handleUserNotFound(ex: ProductCreationException): ResponseEntity<ErrorMessageModel> {
+    @ExceptionHandler(
+        value = [
+            ProductCreationException::class,
+            ProductUpdateException::class,
+            ElementNotFoundException::class,
+        ],
+    )
+    fun handleUserNotFound(ex: RuntimeException): ResponseEntity<ErrorMessageModel> {
```
da-nyee, line 38: Thanks for applying the feedback 😄

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
```diff
@@ -0,0 +1,83 @@
...
+                "status" to "ADDED",
+                "created_at" to now,
+            )
+        return insert.executeAndReturnKey(productMap).toLong()
+    }
+
+    fun getTop5ActiveUsers(): List<ActiveUsersResponse> {
+        val sql =
+            """
+            SELECT DISTINCT m.ID, m.NAME, m.EMAIL FROM MEMBERS m
```
da-nyee, line 75: 👍

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
da-nyee: That's right — you understand the differences between the joins well 👍

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
```diff
@@ -0,0 +1,83 @@
...
+            JOIN CARTS c ON m.id = c.MEMBER_ID
+            JOIN CART_HISTORY ch ON ch.CART_ID = c.ID
+            WHERE ch.CREATED_AT >= CURRENT_DATE - INTERVAL '7' DAY
+            ORDER BY m.ID
```
da-nyee, line 79: 👍

src/main/kotlin/ecommerce/controller/AdminProductController.kt
da-nyee: > 204 (No Content) is typically used when the server successfully processes the request but doesn’t return any content in the response body.

Okay, I totally agree with your point 👍
I think it'd be better to stick with the current status.

src/main/kotlin/ecommerce/configuration/AdminOnlyResolver.kt
da-nyee: Well, they do seem quite similar, but they handle different annotations.
So I think the current structure works well — if you were to merge them, it might end up delegating too many responsibilities to a single class, which could violate the Single Responsibility Principle (SRP).

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
da-nyee: Nice — that index looks solid 👍

da-nyee: Just a small suggestion — have you heard of covering indexes?
It might be worth looking into ☺️

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
da-nyee: Thanks for the detailed explanation 👍

Totally agree that setting up `@BeforeEach` with proper cleanup and authentication can be tricky at first.
The current setup is DB-dependent. That said, I don't think it's a big deal if the environments were separated (e.g., prod / dev / test).
Data consistency should be relatively safe as long as the tests rely solely on a dedicated test DB (e.g., H2) — especially when the DB is used only for testing purposes.

As for the date-sensitive tests in `AdminStatsControllerTest`, you're absolutely right — once those hardcoded dates become outdated, the tests might fail.
Maybe you could replace the fixed dates with something relative to make them more robust, like `NOW()` plus or minus a few days.
That way, the tests won't break over time and will still be reliable regardless of when they run.

da-nyee (changes requested):
Hi Vito! 🙌

I think the mission is just about ready to wrap up, but I'll leave it as a request for changes for now — I've gone through all your comments and left some replies.
Once everything's clear, feel free to request a review again. I'll go ahead and merge it then.

If you have any other questions, just let me know — happy to help ☺️

src/main/kotlin/ecommerce/repository/AdminStatsRepository.kt
VittorioDeMarzi: Thanks for the hint! I’ll definitely have a look into covering indexes 😊

src/test/kotlin/ecommerce/endToEnd/AdminControllerTest.kt
VittorioDeMarzi: Got it, that makes perfect sense 👍
For the future I’ll make sure the tests run against a dedicated test DB so data consistency won’t be an issue.
For the date-sensitive tests, I’ll switch to using relative dates instead of hardcoded ones so they remain stable over time.
Thanks for the tip! 😊

VittorioDeMarzi:
Thanks a lot for the tips, I really appreciate them and will make good use of them 👍
From tomorrow, I can fully shift my focus to the next mission and dive into working with Data JPA 😄🥳

da-nyee (approved):
Hi Vito! 🙌

Everything's looking good, so I'll go ahead and wrap up the mission — congratulations 🎉
I really hope the feedback has been helpful and that you've gotten a lot out of it.

If you have any further questions, feel free to let me know — I'd be happy to help ☺️

PR #26, opened 2025-07-26
[spring-ecommerce-product step2] submit Petra's mission

pebencze:
Hello @Hyeon9mak! 

Hope you are doing well! I have completed step 2 with the following new features:

- **Step 2-1**: Added input validation for product name (length, allowed special characters), price (greater than 0), and image URL (HTTP/HTTPS).
- **Step 2-2**: Implemented user authentication with JWT tokens, allowing user registration, login, and access control based on role (Admin/Member).
- **Step 2-3**: Enabled cart functionality where users can add/remove products, with proper authentication via JWT.
- **Step 2-4**: Added admin statistics endpoints to retrieve top 5 most added products and active members in the last 7 days.

For more details, you can also take a look at my README. There were so many new concepts and features in this mission that I had to implement in a limited time. I think I made progress in understanding how DTOs, Repository, Service, Controller, ControllerAdvice and MvcConfiguration work together. However, I had a hard time testing. I was also experimenting with jdbcClient instead of jdbcTemplate, which might create a little inconsistency between step2 and 3. The most difficult parts were the queries and the decision on the DTO return types for the cart items, since I have to return the product details.
- Q1: I would appreciate if you could take a look at my DTOs and my repository. Are the method arguments there and the return types okay?
- Q2: And in general how responsibility is split between Controller, Service and Repo.

I have also left a few considerations in my README. I will do research on them.

Thanks a lot for all the effort you put into the feedbacks!

Best,
Petra


src/main/kotlin/ecommerce/config/AdminAuthInterceptor.kt
```diff
@@ -0,0 +1,26 @@
...
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        handler: Any
+    ): Boolean {
+        val token = authExtractor.extract(request)
+        if (token.isEmpty()) throw UnauthorizedException()
+        val admin = authService.findMember(token)
+        if (admin.role != "admin") throw UnauthorizedException()
+        return true
+    }
```
Hyeon9mak: What is the different of UNAUTHORIZED and FORBIDDEN?

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+    @ExceptionHandler(UnauthorizedException::class)
+    fun handleUnauthorized(e: UnauthorizedException): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(e.message)
+    }
+
+    @ExceptionHandler(InternalServerErrorException::class)
+    fun handleInternalError(e: InternalServerErrorException): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body(e.message)
+    }
+}
```
Hyeon9mak: How are all the other unexpected exceptions being handled?
Wouldn't it be better to have an additional handler to manage them in one place? 😉

src/main/kotlin/ecommerce/repository/CartRepository.kt
```diff
@@ -0,0 +1,201 @@
...
+class CartRepository(private val jdbcClient: JdbcClient) {
+    fun addItemToCart(
+        productId: Long,
+        productQuantity: Long,
+        cartId: Long,
+    ): CartItemDto {
+        val updateSql = """
+        UPDATE cart_items
+        SET quantity = quantity + ?
+        WHERE product_id = ? AND cart_id = ?
```
Hyeon9mak: It names `CartRepository`, but it handles `CartItem`.
What's different of two words?

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,59 @@
...
+    fun login(request: TokenRequest): TokenResponse {
+        if (!memberRepository.existsByEmail(request.email)) {
+            throw NotFoundException("No account with email exists")
+        }
+        val member = Member(email = request.email, password = request.password)
+        if (memberRepository.findMember(member) == null) {
+            throw UnauthorizedException("Incorrect Password")
+        }
+        return createToken(request)
+    }
```
Hyeon9mak: It seems like 2 queries are being triggered.
How about asking the created Member instead?

```kotlin
member.validatePassword(request.password)
```

we already learned about "send message to object" 😉

Hyeon9mak (changes requested):
Hi Petra,
Hope you're having a relaxing weekend!

I noticed that some of the tests are currently failing when run all at once.
Most of them seem to be due to port conflicts, but quite a few are failing because of issues within the test logic itself.

It would be great if you could go through and fix these issues before requesting a review.

image

At HeroTechCourse, I truly hope that our crew isn’t just rushing to finish assignments, but instead taking the time to grow as solid engineers.
Please take this as an opportunity to work with your teammates to identify the root causes and find solutions together.
And once you've done that — I’d love for you to explain to me how you resolved them! 😉 

With that in mind, I’ve left only a few review comments this time to help you stay focused on the debugging process. 😄

src/main/kotlin/ecommerce/config/AdminAuthInterceptor.kt
pebencze: Thank you for the reminder! In cases where the token is valid, but the user is e.g. not admin and not allowed to make certain requests, 403 Forbidden is more suited. If the credentials are invalid, so the user cannot be authorized, 401 Unauthorized is more suited.

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
pebencze: Makes a lot of sense! Just added it. Thanks.

src/main/kotlin/ecommerce/repository/CartRepository.kt
pebencze: Well, it handles both the Cart (e.g. create or find a user's cart) and the CartItems (e.g. add product to cart) database tables. I can also split it into two repositories, if you'd like, but for now I thought this would be enough, because there is only the findOrCreateCartId() method that could be extracted to a separate repo. Now as I reflect, it would be better to separate them, but if the cart repo does not grow, it is okay to keep it like this. What do you think?

src/main/kotlin/ecommerce/service/AuthService.kt
pebencze: Insightful. Hope the new implementation follows this advice.

pebencze:
Hello GuMak! 

I have resolved the mentioned issues and made quite a few additional changes.

I have made the tests work and pass, polished them and extracted some logic into helper methods. Concerning the project itself, I have implemented 3 of your suggestions and done following changes:
- the Interceptor works both for admins and users: it extracts the token and sets the "email" attribute inside the request
- there are 2 custom annotations now: @LoginMember and @Admin, that check if the member exists inside the DB and for admin it additionally validates the role
- I have also updated my removeItem and addItem methods inside the cart repository, so that the statistics can be calculated

During this process I have realized that I still don't feel confident when it comes to splitting the responsibility between different components, like: Repository, Controller, Service. Asking the object to validate itself with the password was also quite new to me. How can I split responsibility better?
I also have a StatRepository, but no separate table for stats... Since I use the cart_items table, shall I put those two functions into the CartRepository? Is there a rule or best practice for that decision?
Another point is: where should I use DTOs and where domain objects/models?

Would appreciate if you had time to review my project. Thanks in advance!

All the best,

Petra

src/main/kotlin/ecommerce/annotation/AdminArgumentResolver.kt
```diff
@@ -0,0 +1,46 @@
...
+        try {
+            member = authService.findMember(email.toString())
+        } catch (e: NotFoundException) {
+            throw UnauthorizedException()
+        }
+
+        if (member.role != "admin") {
+            throw ForbiddenException()
+        }
+
```
Hyeon9mak: 👍 
And more, admin role is not a just string! so we can declare role enum class 😉 

And more! u can ask to `AuthService`.

```kotlin
    override fun resolveArgument(
        parameter: MethodParameter,
        mavContainer: ModelAndViewContainer?,
        webRequest: NativeWebRequest,
        binderFactory: WebDataBinderFactory?,
    ): Any {
        val request = webRequest.getNativeRequest(HttpServletRequest::class.java) ?: throw UnauthorizedException()
        val email = request.getAttribute("email").toString()
        return authService.findAdminMember(email)
    }
```

All security-related responsibilities can be delegated to the `AuthService`, allowing the Resolver to focus solely on parameter passing.
This will likely make writing test code much easier as well!

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,42 @@
...
+    @ExceptionHandler(ForbiddenException::class)
+    fun handleInternalError(e: ForbiddenException): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.FORBIDDEN).body(e.message)
+    }
+
+    @ExceptionHandler(Exception::class)
+    fun handleGenericException(e: Exception): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR)
+            .body("An unexpected error occurred: ${e.localizedMessage ?: "Unknown error"}")
+    }
```
Hyeon9mak: Nice 👍 
we can't handle unexpected error, it's right to return internal server error status code.

And, there is a probability of sensitive info in error message.
So It's better not to deliver the message.

src/main/kotlin/ecommerce/exception/NotFoundException.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.exception
+
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.ResponseStatus
+
+@ResponseStatus(HttpStatus.NOT_FOUND)
+class NotFoundException(override val message: String? = null) : RuntimeException()
```
Hyeon9mak, line 7: what about declare kind message for user? 😉

src/main/kotlin/ecommerce/repository/CartRepository.kt
```diff
@@ -0,0 +1,206 @@
...
+    ): CartItemDto {
+        val updateSql =
+            """
+            UPDATE cart_items
+            SET quantity = quantity + ?, created_at = CURRENT_TIMESTAMP
+            WHERE product_id = ? AND cart_id = ?
+            """.trimIndent()
+
+        val insertSql =
+            """
```
Hyeon9mak: >  it handles both the Cart (e.g. create or find a user's cart) and the CartItems (e.g. add product to cart) database tables. I can also split it into two repositories, if you'd like, but for now I thought this would be enough, because there is only the findOrCreateCartId() method that could be extracted to a separate repo. Now as I reflect, it would be better to separate them, but if the cart repo does not grow, it is okay to keep it like this. What do you think?

That's what you meant!
The term repository originates from Domain-Driven Design (DDD), based on the idea of treating database persistence objects as if they were regular Java objects.
So it can be helpful to think of a repository in the same way you think of a Collection, such as a List or Set.

For example, it would feel unnatural if a call like `List<Cart>.get()` were to return a CartItem.
In the same way, the return values of a repository should align with the aggregate or entity it represents. 😉

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    fun login(request: TokenRequest): TokenResponse {
+        val member =
+            memberRepository.findByEmail(request.email)
+                ?: throw UnauthorizedException("No account with email exists")
+        member.validatePassword(request.password)
+        return createToken(request)
+    }
```
Hyeon9mak, line 57: > Hope the new implementation follows this advice. 

💯

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,26 @@
...
+    @field:NotBlank(message = "Field cannot be blank")
+    @field:Pattern(regexp = "^(user|admin)", message = "Invalid role")
+    val role: String = "user",
+) {
+    fun validatePassword(password: String) {
+        if (this.password != password) {
+            throw UnauthorizedException("Incorrect password")
+        }
+    }
+}
```
Hyeon9mak: 99 of 100.
1 is test code 😉

src/main/kotlin/ecommerce/annotation/AdminArgumentResolver.kt
```diff
@@ -0,0 +1,46 @@
...
+    override fun resolveArgument(
+        parameter: MethodParameter,
+        mavContainer: ModelAndViewContainer?,
+        webRequest: NativeWebRequest,
+        binderFactory: WebDataBinderFactory?,
+    ): Any {
+        lateinit var member: MemberDto
+        val request =
+            webRequest.getNativeRequest(HttpServletRequest::class.java)
+                ?: throw UnauthorizedException()
```
Hyeon9mak: it seems not auth error! 😉

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -1,60 +1,58 @@
...
 class ProductController(private val productRepository: ProductRepository) {
     @PostMapping("/products")
-    @ResponseBody
     fun create(
-        @RequestBody product: Product,
+        @RequestBody @Valid product: Product,
     ): ResponseEntity<Unit> {
+        if (productRepository.existsByName(product.name)) {
+            throw IllegalArgumentException("Product with name ${product.name} already exists")
+        }
```
Hyeon9mak: I think it's business logic, not HTTP protocol logic!
and it's hard to declare test code.
What do u think? 😄

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
```diff
@@ -0,0 +1,30 @@
...
+        resolvers.add(LoginMemberArgumentResolver(authService))
+        resolvers.add(AdminArgumentResolver(authService))
+    }
+
+    override fun addInterceptors(registry: InterceptorRegistry) {
+        registry.addInterceptor(AuthInterceptor(authExtractor, jwtTokenProvider))
+            .addPathPatterns("/admin/**")
+            .addPathPatterns("/api/**")
+            .excludePathPatterns("/api/members/login", "/api/members/register", "/api/members/me/token")
+    }
```
Hyeon9mak: Maybe... `/api/members/me/token` no need exclude pattern..? 🤔

src/main/kotlin/ecommerce/dto/TokenRequest.kt
```diff
@@ -0,0 +1,14 @@
...
+import jakarta.validation.constraints.Size
+
+class TokenRequest(
+    @field:NotBlank(message = "Field cannot be blank")
+    @field:Email(message = "Invalid email")
+    val email: String,
+    @field:NotBlank(message = "Field cannot be blank")
+    @field:Size(min = 8, max = 30, message = "Invalid password: must between 8 and 30 characters")
+    val password: String,
+)
```
Hyeon9mak, line 14: Request/Response DTOs and Domain Objects are used in the Presentation Layer and Domain Layer, respectively.
In your opinion, which layer is more appropriate for handling validation?

src/main/resources/application.properties
```diff
@@ -6,4 +6,7 @@ spring.jpa.defer-datasource-initialization=true
 spring.sql.init.mode=always
 spring.sql.init.schema-locations=classpath:schema.sql
 spring.sql.init.data-locations=classpath:data.sql
+security.jwt.token.secret-key=your-very-secret-key-super-super-long-special-key
+security.jwt.token.expire-length=3600000
+
```
Hyeon9mak: seconds? minutes? hours? days? months?
we can rename with time unit or append comment 😉

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -0,0 +1,32 @@
...
+        if (other == null || javaClass != other.javaClass) return false
+
+        other as Product
+
+        return id != null && id == other.id
+    }
+
+    override fun hashCode(): Int {
+        return id?.hashCode() ?: System.identityHashCode(this)
+    }
```
Hyeon9mak: 👍 Also Member have `id` 😄

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,54 @@
...
+        return MemberDto(member.id!!, member.email, member.role)
+    }
+
+    fun findMemberByToken(token: String): MemberDto {
+        if (!jwtTokenProvider.validateToken(token)) {
+            throw UnauthorizedException("Invalid token")
+        }
+        val payload = jwtTokenProvider.getPayload(token)
+        return findMember(payload)
+    }
```
Hyeon9mak: how about ask to jwtTokenProvider?
`if (jwtTokenProvider.isInvalid(token)) { ... }`

or throw exception it self in `getPayload` method.

src/main/kotlin/ecommerce/service/CartService.kt
```diff
@@ -0,0 +1,43 @@
...
+    fun deleteItem(
+        memberId: Long,
+        request: CartItemRequest,
+    ): CartItemDto? {
+        val cartId = cartRepository.findOrCreateCartId(memberId)
+        if (!productRepository.existsById(request.productId)) {
+            throw NotFoundException("Product with ${request.productId} not found")
+        }
+        return cartRepository.removeItemFromCart(request.productId, request.quantity, cartId)
+    }
```
Hyeon9mak: > Asking the object to validate itself with the password was also quite new to me. How can I split responsibility better?

Terms like Controller, Service, Repository, and Domain are all expressions of intention to draw boundaries.
These boundaries are based on separation of concerns. But why were they divided this way?
It’s because we have a desire to group and modify our code by purpose—whether it’s HTTP, business logic, or database (persistence).
(Of course, the origins of these terms may differ, but let’s assume this to simplify understanding.)

Controller is part of the presentation layer and handles HTTP requests.
You can think of it as a translator between our business logic and the HTTP layer.
Everything related to the HTTP protocol should be managed by the Controller, and anything business-related should be passed to the Service.

Service is responsible for coordinating the flow.
Business logic itself should be delegated to the Domain layer.
So the Service layer simply tells the Domain and Repository what to do based on the request it receives from the Controller.
Instead of calling getters from the Domain, it sends commands and waits for the results, then passes them to the Repository when needed.

Domain exists to encapsulate and execute business logic.
This makes the Domain layer less dependent on external technologies, which in turn makes it easier to test.
It’s also the layer where we embed business rules, so that even other developers can understand the business just by looking at the Domain.

Repository is used to represent the persistence (DB) layer in the JVM world.
I believe you’re already quite familiar with it, so I’ll skip further explanation.

Now, does this give you a clearer picture of the overall structure? 😄

src/main/kotlin/ecommerce/dto/CartItemDto.kt
```diff
@@ -0,0 +1,9 @@
+package ecommerce.dto
+
+class CartItemDto(
+    var quantity: Long,
+    val productId: Long,
+    val productName: String,
+    val productPrice: Double,
+    val productImageUrl: String,
+)
```
Hyeon9mak: > where should I use DTOs and where domain objects/models?

As the name suggests, a DTO (Data Transfer Object) is used to transfer data between different layers in your application.
These layers typically include Controller, Service, Domain, and Repository, as we discussed earlier.

Let’s consider an example where the Controller layer frequently changes.
If the same object is shared across Controller, Service, Domain, and Repository, then a small change in the Controller might ripple through all other layers.
To reduce this coupling and make the codebase more resilient to change, we use DTOs to separate the layers.
The Controller can extract values from the DTO and pass them to the Service, or the Service can extract data from the DTO and pass it down to the Domain or Repository.
This minimizes how many layers are affected by a change in the Controller.

Because of this nature, DTOs should not contain any business logic such as “the name must be 3–8 characters long.”
DTOs are meant to be lightweight, disposable objects used strictly for data transfer between layers.
They don’t need to include validation logic or unit tests.
If you feel like adding business logic into a DTO seems natural, that’s usually a sign that the object is no longer a DTO—it’s a Domain model.

To take the idea even further: in some cases, even an entity that contains an id field can be thought of as a DTO. (Absolutely NOT)
Why? Because the id is primarily a persistence concern—not a business concern.
So in the Repository layer, we can receive a pure Domain model, convert it to an entity (with id) for database use,
and then convert it back to a Domain model when returning the result.
This approach makes your Domain layer independent of database structures and much easier to test.

The root of confusion often comes from thinking that DTO, Domain model, and Entity must always match 1:1:1.
But that’s not the case:

- A DTO is simply a "truck" that moves data between layers.
- A Domain model is a collection of business behaviors (methods), with no concern for how data is stored.
- An Entity is a representation of data in the database and usually includes an id.

Here’s how these roles might play out in practice:

1. MemberRegisterRequest(DTO) → RegisterMember(DM) → MemberEntity → RegisteredMember(DM) → MemberRegisteredResponse(DTO)
2. MemberLoadRequest(DTO) → LoadMember(DM) → MemberEntity → LoadedMember(DM) → MemberLoadedResponse(DTO)
3. MemberDeleteRequest(DTO) → X → MemberEntity
4. MemberLoginRequest(DTO) → MemberEntity → FoundedMember(DM) → MemberPasswordValidator(DM) → MemberEventAgreeChecker(DM) -> MemberPhoneNumberAliveChecker(DM) → ... → MemberLoginResponse(DTO)

Hyeon9mak (changes requested):
Hi Petra! Sorry for late 🙏 

All your tests are passing well—great job on making the changes!
I know it must have taken quite some effort. 👍 👍 

Understanding the concept of layers in an application can be tricky, right?
You often hear terms like Controller, Service, Domain, and Repository,
but rarely do people explain where they come from or what pros and cons they bring in real-world use. 👻 

It’s completely normal to find it challenging—especially when you haven’t had much experience to see these ideas play out.
But if you keep consciously thinking about it and practicing how to apply these layers, you’ll naturally gain more clarity over time. 💪 

I’ve left a few comments along with detailed answers to your questions.
Feel free to ask anything else that comes to mind! 😄

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
pebencze: Related to this, is it a good practice to say that a user with the same email already exists, or would I just return a general error message, like "Email or password are incorrect", when I throw Unauthorized exceptions inside the register or login function.

src/main/kotlin/ecommerce/config/WebMvcConfiguration.kt
pebencze: True. It was kind of a method for learning purposes, but I forgot to refactor it.

src/main/kotlin/ecommerce/dto/TokenRequest.kt
pebencze: After reflecting on it, I would rather validate on the Presentation Layer, because I could catch the error as soon as possible and would not make unnecessary queries... 

If I don't validate on all layers, could it introduce bugs or security weaknesses to my application?

src/main/kotlin/ecommerce/model/Product.kt
pebencze: I have overridden equals and hashcode, but I actually don't use them and did not see an opportunity yet. Will it be useful if I start using JPA?

src/main/kotlin/ecommerce/service/AuthService.kt
pebencze: I think it is fine like this, because it separates responsibilities. The drawback is that you could forget to validate the token and get the payload of an expired token...

src/main/kotlin/ecommerce/dto/TokenRequest.kt
pebencze: New thoughts: it is best to validate business logic in the domain models within the service layer, but jakarta validation for simple input is done in the controller. The advantage of it is what I have mentioned above. Since emails don't really have a special format, I think it is okay to validate them here and just checks for duplicates in side the service. 
However, passwords could be validated on several layers and especially inside the domain model. For the scope of this project I will just stick to jackarta, because I did not yet implement a special password policy that would be crucial to my domain.

pebencze:
Hello GuMak! 

Thank you for the comments, especially for clarifying the underlying concepts of this mission. I think I have learnt a lot today trying to implement your feedback, discussing with my colleagues and experiment with different approaches when it comes to the separation of responsibilities.

I would really appreciate if you could look at my new commits. Since I am starting a new project tomorrow based on this code, I would be grateful about some feedback.

At the end of my README, you can find the refactoring points. I have implemented the ones that are checked, I decided not to implement some of them and some of them I would still like to incorporate either into this or into the next mission. E.g. I would like to refactor the tests using init.sql and @Sql.

By now I have read and discussed  a lot about layered architecture and I have decided to go with the following approach: for the scope of this mission, entity and model represent the same thing - they both hold IDs, and e.g. inside the Product and Member domains we have business logic. DTOs are mainly Request and Response data holders and Request data holders contain jakarta validation to detect input error as soon as possible. I think I could follow this idea quite well for Product and Member related classes. 

I have also tried implementing Entities that do not hold any business logic, and Domain Models that do not hold IDs, but it would have taken a lot of time for no significant advantages for this mission. Also, in my next project, I will be implementing JPA and would like to deepen my knowledge there.

For the Cart related layers: I did not have time to refactor them yet. I would have refactored them using the CartItem and Cart domain models. E.g. inside the CartService I could separate the logic. Instead of making joined requests. 

I have added 'name' to the Member class and database, as well, and implemented the feedback from your comments.

Please let me know what you think about the refactored code and if I can proceed to the next mission with this codebase. I will definitely keep in mind what I have learned throughout this project.

Thanks in advance!

Best,

Petra

src/main/kotlin/ecommerce/annotation/AdminArgumentResolver.kt
pebencze: How is the new enum class?

README.md
```diff
@@ -262,4 +262,78 @@ of Long types Value Class types like ProductId, or should it take a CartRequestD
...
+  - not a priority now, logic is separated, so I will keep the current implementation
+- [ ] try to refactor structure according to these principles:
+  - [ ] everything related to the HTTP protocol should be managed by the Controller, and anything business-related should be passed to the Service
+    - A DTO is simply a "truck" that moves data between layers.
+    - A Domain model is a collection of business behaviors (methods), with no concern for how data is stored.
+    - An Entity is a representation of data in the database and usually includes an id.
+  - e.g.: MemberRegisterRequest(DTO) → RegisterMember(DM) → MemberEntity → RegisteredMember(DM) → MemberRegisteredResponse(DTO)
+- [ ] test password validation after refactoring
+- [x] override equals() and hashcode() for Member
 -
```
Hyeon9mak, line 304: When you get the chance, I really encourage you to finish everything—it’ll become a strong asset for your future growth 😄

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -37,6 +38,6 @@ class GlobalExceptionHandler {
     @ExceptionHandler(Exception::class)
     fun handleGenericException(e: Exception): ResponseEntity<String> {
         return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR)
-            .body("An unexpected error occurred: ${e.localizedMessage ?: "Unknown error"}")
+            .body("An unexpected error occurred.")
     }
```
Hyeon9mak, line 42: 👍

src/main/kotlin/ecommerce/dto/AllCartItemsResponse.kt
```diff
@@ -0,0 +1,6 @@
+package ecommerce.dto
+
+class AllCartItemsResponse(
+    val cartId: Long,
+    val items: List<CartItemResponse> = emptyList(),
+)
```
Hyeon9mak, line 6: one more tip!
DTOs are typically a great fit for data class,
since they often benefit from automatically implemented `toString`, `equals`, and `hashCode` methods. 😄

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.MethodArgumentNotValidException
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
+
+@ControllerAdvice
+class GlobalExceptionHandler {
+    @ExceptionHandler(MethodArgumentNotValidException::class)
```
Hyeon9mak, line 11: > Related to this, is it a good practice to say that a user with the same email already exists, or would I just return a general error message, like "Email or password are incorrect", when I throw Unauthorized exceptions inside the register or login function.

Email duplication isn't a serious security risk — it's simply a check to see if the email already exists during signup. So it's better to explain the reason clearly to the user.

src/main/kotlin/ecommerce/model/Product.kt
```diff
@@ -1,10 +1,10 @@
...
-data class Product(
-    var id: Long? = null,
-    var name: String = "",
-    var price: Double = 0.0,
-    var imageUrl: String = "",
+class Product(
+    var id: Long?,
+    var name: String,
+    var price: Double,
+    var imageUrl: String,
```
Hyeon9mak, line 7: > I have overridden equals and hashcode, but I actually don't use them and did not see an opportunity yet. Will it be useful if I start using JPA?

ID and entity are not concepts exclusive to JPA — they are much broader. By overriding equals and hashCode, you're explicitly conveying these concepts to fellow developers!

Hyeon9mak (approved):
Hi Petra! Great job reflecting the feedback!

I’ve left some comments answering your questions.
As you move on to the next mission, where you'll start using JPA, I expect the concepts and terms we've discussed in this mission to become clearer. 😄 

Don't feel rushed—take your time to understand the ideas step by step.

I hope the feedback I’ve given so far has been helpful.
Approved!

PR #29, opened 2025-07-28
[Spring-Ecommerce-Product Step 2] Submit Farhana's Mission

farhanaahmed:
# PR: Spring E-commerce - Implement Validation, Authentication, Cart, and Admin Analytics (Step 2-1 to 2-4)

## Step 2-1: Product Validation

- Added validation rules for Product creation and update:

- Product name: max 15 characters, allowed special characters only, must be unique

- Product price: must be greater than 0

- Image URL: must start with http:// or https://

- Invalid data now returns clear error messages via @ControllerAdvice.

## Step 2-2: Member Registration and Login

- Member registration with email and password.

- Login returns a JWT token upon valid credentials.

- Token to be used for protected endpoints.

## Step 2-3: Member Cart Functionality

- Members can:

- Add products to their cart (token required)

- View their cart

- Remove products from their cart

## Step 2-4: Admin Analytics

- Endpoint to get Top 5 most added products in the last 30 days:

- Includes product name, count, and the most recent added time

- Endpoint to get members who added products to cart in the last 7 days:

- Includes member ID, name, and email

- Unique members only, even if multiple items were added

- Access to analytics endpoints restricted to ADMIN users via interceptor

## Testing Notes

I tried to follow Test-Driven Development (TDD) as much as I could throughout these steps.
As I’m still learning Spring Boot, there were some situations (especially with security, interceptors, and complex JDBC logic) where I wasn’t sure how to approach testing properly.

I’ve written unit tests where I was confident, and I’d appreciate your feedback or pointers on:

- How to test JWT-protected endpoints effectively

- Best practices for writing tests around interceptors and analytics queries

As I’m learning and applying different testing strategies like End-to-End, Slice, Web Layer, Service Layer, and Integration testing, I want to make sure I'm focusing on the most important aspects.

In your experience, what kinds of test cases should I absolutely include when working on a Spring Boot project?
Are there any types of tests you consider non-negotiable for production-quality code, or any testing best practices you think I should follow consistently?

I’d really appreciate your guidance on how to improve my test coverage and approach. Thanks in advance. 

Thank you for reviewing! 🌱 
Looking forward to your valuable feedback.


src/main/kotlin/ecommerce/config/WebConfig.kt
```diff
@@ -0,0 +1,28 @@
...
+import org.springframework.web.method.support.HandlerMethodArgumentResolver
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer
+
+@Configuration
+class WebConfig(
+    private val authInterceptor: AuthInterceptor,
+    private val loginMemberArgumentResolver: LoginMemberArgumentResolver,
+    private val adminInterceptor: AdminInterceptor,
+) : WebMvcConfigurer {
```
Hyeon9mak: What's different of AuthInterceptor & AdminInterceptor?
I think u can give more specific name to interceptors 😄

src/main/kotlin/ecommerce/controller/AdminController.kt
```diff
@@ -0,0 +1,27 @@
...
+    fun findTop5ProductsInLast30Days(): ResponseEntity<List<TopProductStatResponse>> {
+        val topProducts = cartService.findTop5ProductsInLast30Days()
+        return ResponseEntity.ok(topProducts)
+    }
+
+    @GetMapping("/cartactivity")
+    fun findMembersWithCartActivityInLast7Days(): ResponseEntity<List<MemberResponse>> {
+        val activeMembers = cartService.findMembersWithCartActivityInLast7Days()
+        return ResponseEntity.ok(activeMembers)
+    }
```
Hyeon9mak: U can give more readable URIs for them 😄 

> 2.2.3. Use hyphens (-) to improve the readability of URIs
> To make your URIs easy for people to scan and interpret, use the hyphen (-) character to improve the readability of names in long-path segments.
>
https://restfulapi.net/resource-naming/?utm_source=chatgpt.com

src/main/kotlin/ecommerce/dto/ProductRequest.kt
```diff
@@ -0,0 +1,28 @@
...
+    @field:Pattern(
+        regexp = "^(http://|https://).*$",
+        message = "Image URL must start with http:// or https://",
+    )
+    val imageUrl: String,
+) {
+    fun toProduct(): Product {
+        return Product(name = name, price = price, imageUrl = imageUrl)
+    }
+}
```
Hyeon9mak, line 28: Request/Response DTOs and Domain Objects are used in the Presentation Layer and Domain Layer, respectively.
In your opinion, Farhana, which layer is more appropriate for handling validation?

Of course, there's no absolute right answer here.
Performing validation in both layers is the safest approach, but it can also lead to redundant checks.
I'm curious to hear your thoughts.

src/main/kotlin/ecommerce/handler/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+data class ValidationErrorResponse(val errors: List<ValidationError>)
+
+@ControllerAdvice
+class GlobalExceptionHandler {
+    @ExceptionHandler(NoSuchElementException::class)
+    fun handleNoSuchElementException(e: NoSuchElementException): ResponseEntity<Void> {
+        println("NoSuchElementException occurred: ${e.message}")
+        return ResponseEntity.notFound().build()
+    }
+
```
Hyeon9mak: What's different of print & log ? 😉 
It might be a good idea to take some time to learn about the basic logging mechanisms provided by Spring. 😄

Hyeon9mak: Shall we try to recall something together.
When you encountered a 404 NOT FOUND error page, what was the first thing that came to your mind?

- 'I guess the data doesn’t exist...'
- 'Did I type the wrong URL?'

Depending on that, we might need to reconsider which error code to return. 😄

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
```diff
@@ -0,0 +1,11 @@
+package ecommerce.infrastructure
+
+import jakarta.servlet.http.HttpServletRequest
+
+interface AuthorizationExtractor<T> {
+    fun extract(request: HttpServletRequest): T
+
+    companion object {
+        const val AUTHORIZATION = "Authorization"
+    }
```
Hyeon9mak: same thing in `org.springframework.http.HttpHeaders.AUTHORIZATION` 😉

src/main/kotlin/ecommerce/infrastructure/BearerAuthorizationExtractor.kt
```diff
@@ -0,0 +1,28 @@
+package ecommerce.infrastructure
+
+import jakarta.servlet.http.HttpServletRequest
+
+class BearerAuthorizationExtractor : AuthorizationExtractor<String> {
+    override fun extract(request: HttpServletRequest): String {
+        val headers = request.getHeaders(AuthorizationExtractor.AUTHORIZATION)
+        while (headers.hasMoreElements()) {
+            val value = headers.nextElement()
+            if (value.lowercase().startsWith(BEARER_TYPE.lowercase())) {
```
Hyeon9mak: I think not using it anymore.

src/main/kotlin/ecommerce/infrastructure/JWTProvider.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.jsonwebtoken.JwtException
+import io.jsonwebtoken.Jwts
+import io.jsonwebtoken.SignatureAlgorithm
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.stereotype.Component
+import java.util.Date
+
+@Component
+class JWTProvider {
+    @Value("\${security.jwt.token.secret-key}")
```
Hyeon9mak: Hide secret key! nice 💯 👍

src/main/resources/application.properties
```diff
@@ -8,4 +8,7 @@ spring.sql.init.mode=always
 spring.jpa.hibernate.ddl-auto=none
 spring.sql.init.schema-locations=classpath:schema.sql
 spring.sql.init.data-locations=classpath:data.sql
-server.port=8080
\ No newline at end of file
+server.port=8080
+server.error.whitelabel.enabled=false
+security.jwt.token.secret-key=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIiLCJuYW1lIjoiSm9obiBEb2UiLCJpYXQiOjE1MTYyMzkwMjJ9.ih1aovtQShabQ7l0cINw4k1fagApg3qLWiB8Kt59Lno
+security.jwt.token.expire-length=3600000
```
Hyeon9mak: what is expire-length's unit? second? minute? hours? days?
u can add comment or rename it 😄

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -25,80 +25,275 @@ class ProductControllerTest {
...
+            RestAssured.given()
+                .contentType(ContentType.JSON)
+                .body(invalidProduct)
+                .put("/products/$productId")
+                .then()
+                .extract()
+
+        Assertions.assertThat(response.statusCode()).isEqualTo(HttpStatus.BAD_REQUEST.value())
+        Assertions.assertThat(response.asString()).contains("Product name must not be blank", "name")
+    }
```
Hyeon9mak, line 291: 😉 👍 
What matters most in testing, in my view, is identifying exactly what we want to validate.
From that standpoint, this test does a great job!
It clearly checks the response status code and confirms that the expected message is present.

src/test/kotlin/ecommerce/controller/AdminControllerTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        val memberResponse = MemberResponse(1L, "user@example.com", name = "John Doe", role = UserRole.USER.toString())
+        `when`(authService.findMemberByToken(token)).thenReturn(memberResponse)
+
+        mockMvc.perform(
+            get("/admin/topproducts")
+                .header("Authorization", "Bearer $token"),
+        ).andExpect {
+            status().isUnauthorized
+        }
+    }
```
Hyeon9mak: Just like the confidence it shows, this is a well-written and solid test 👍
By mocking the service layer, the controller layer is tested thoroughly and in isolation.

src/test/kotlin/ecommerce/controller/AdminControllerTest.kt
```diff
@@ -0,0 +1,131 @@
...
+    fun `should return status code 401 for cartactivity endpoint when User role is not Admin`() {
+        val memberResponse = MemberResponse(1L, "user@example.com", name = "John Doe", role = UserRole.USER.toString())
+        `when`(authService.findMemberByToken(token)).thenReturn(memberResponse)
+
+        mockMvc.perform(
+            get("/admin/cartactivity")
+                .header("Authorization", "Bearer $token"),
+        ).andExpect {
+            status().isUnauthorized
+        }
```
Hyeon9mak: From my personal experience, when time is limited, the two most essential types of tests are end-to-end (E2E) tests and unit tests.

E2E tests are critical from a business perspective — they ensure that the application meets the requirements as expected by stakeholders.

Unit tests, on the other hand, are valuable because they verify the correctness of small, individual components quickly and efficiently. Since they don’t rely on Spring Beans, they don’t require the Spring context to be loaded, resulting in faster execution and allowing for hundreds or even thousands of tests to run in parallel.

Other testing approaches certainly have their value in specific contexts, but based on my experience, these two types offer the highest return on investment. In particular, slice or integration tests are often used when there’s a lack of technical confidence or clarity. Once a deeper understanding of the system is gained, these can often be reduced or even omitted.

Hyeon9mak (changes requested):
Hi Farhana, great work in Step 2!

Especially, lot of test cases and test code are very impressive. 👍 

I left few comments and answer of ur question 😄 
Feel free to ask anytime!

src/main/kotlin/ecommerce/config/WebConfig.kt
farhanaahmed: Thanks for the feedback! You’re right, the interceptor names are too generic.

What each one does: 

- `AuthInterceptor` –  It is used for general authentication, checking if the user is logged in. It checks that the request contains a valid JWT for any authenticated user (member or admin).

- `AdminInterceptor` –  It is used for authorization, ensuring that only users with the `ADMIN` role can access certain endpoints. It runs after the auth check and verifies that the authenticated user has the ADMIN role before it lets the request reach admin-only controllers.

I have renamed  them to `JwtAuthenticationInterceptor` and `AdminRoleCheckInterceptor` to make the purpose clearer. Let me know if you have any other naming suggestions! 😄

src/main/kotlin/ecommerce/controller/AdminController.kt
farhanaahmed: Thanks for the suggestion and the reference! 😄

You're right about using hyphens. It makes the URIs more readable and RESTful.

I have updated the endpoints to something like:

- `/top-products` instead of `/topproducts`
- `/cart-activity` instead of `/cartactivity`
- `/find-member` instead of `/findmember`

Let me know if you have a preferred naming style or anything else I should improve!

src/main/kotlin/ecommerce/dto/ProductRequest.kt
farhanaahmed: That's a great question and I agree, there's no one-size-fits-all solution! 😊

After working through the implementation, I now see the value of layering validations.

I think general input validations (like blank checks or malformed URLs) belong in the Presentation Layer, it helps catch bad user input early and keeps the request clean.

But for critical business rules like ensuring the product name is unique or enforcing minimum price, I’ve moved those into the Domain Layer, so they're always enforced regardless of the entry point (API, batch jobs, etc.).

So my updated approach looks like this:

Presentation Layer: general format validations (e.g., blank, size, pattern)

Domain Layer: business rules and invariants (e.g., uniqueness, pricing rules)

Implementing it this way felt more robust and aligned with clean architecture principles. 
Curious to hear your thoughts on this! 😄

src/main/kotlin/ecommerce/handler/GlobalExceptionHandler.kt
farhanaahmed: Thanks for the nudge. I totally agree.

println only dumps text to stdout but the application’s logging framework (SLF4J/Logback) gives us:

- Log levels (DEBUG, INFO, WARN, ERROR) so we can silence or escalate messages via configuration.
- Structured output (timestamp, thread, MDC) that’s easy to grep/aggregate in production.
- Performance (lazy string concatenation, async appenders).
- Replaced `println` with proper SLF4J logging (`logger.warn`) so we get levels and structured output.

When I see a *404*, the first thought is usually “client asked for a resource that does not exist” which is exactly what handler returns.
I deliberately throw NoSuchElementException from the service layer when an ID is not found, so returning 404 is consistent. If I ever need to distinguish “wrong URL” from “entity missing”, it can be mapped to a different exception (e.g., MethodArgumentNotValidException → 400).So, I kept 404 for `NoSuchElementException`. It matches the “resource not found” scenario. 
But please let me know your perspective about this. 🙏

src/main/kotlin/ecommerce/infrastructure/AuthorizationExtractor.kt
farhanaahmed: Got it! 😄 You're right. I should use `HttpHeaders.AUTHORIZATION` instead of hardcoding the "Authorization" string. It makes the code cleaner and less error-prone. Thanks for pointing that out. I made the changes. 🙌

src/main/kotlin/ecommerce/infrastructure/BearerAuthorizationExtractor.kt
farhanaahmed: You’re right.I switched to `HttpHeaders.AUTHORIZATION` in the last commit.
It looks somehting like this now: 
`val headers = request.getHeaders(HttpHeaders.AUTHORIZATION)`
The companion object in `AuthorizationExtractor` has also been removed, so `AuthorizationExtractor.AUTHORIZATION` is gone.

src/main/resources/application.properties
farhanaahmed: Good point, the unit isn’t obvious.
It’s milliseconds (3600000 ms = 1 hour).
I have renamed the property to make it self-documenting like this:
`security.jwt.token.expire-time-millis=3600000`

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
farhanaahmed: Thanks for the feedback! 🙌
I'm glad that the test’s intent and assertions are clear. I’ll try to keep the same style for the rest of the tests.

src/test/kotlin/ecommerce/controller/AdminControllerTest.kt
farhanaahmed: Thank you! 😄 I’ll keep the same mocking strategy for the other role-guard tests.

src/test/kotlin/ecommerce/controller/AdminControllerTest.kt
farhanaahmed: Thanks for sharing your perspective. It is very pragmatic. I find it very helpful to prioritize test cases.
I’ve tried to follow the same approach as you suggested. With the limited time, I focused on E2E tests that cover every controller endpoint and complemented them with fast unit tests for domain rules. 
I’ll keep prioritizing E2E for business flow coverage and unit tests for fast feedback, and only add slice/integration tests when we truly need to validate DB wiring or edge-case behaviour.

farhanaahmed:
Hello @Hyeon9mak 😄,
Thank you for all the insightful feedback. I’ve applied every suggestion: 

- Renamed the interceptors and switched to kebab-case admin URIs.
- Moved validation responsibilities: DTOs keep syntactic checks (@NotBlank, etc.) while the domain layer now enforces business invariants (price > 0, unique name).
- Clarified the JWT expiration property with the expire-time-millis name and corrected its value.
- Replaced println with SLF4J logging in GlobalExceptionHandler.
- Removed the duplicate AUTHORIZATION constant in favour of HttpHeaders.AUTHORIZATION.
- Continued focusing on E2E and unit tests as you recommended. 

All tests are also passing. Would you kindly take another look when convenient?
I appreciate your time and guidance! 🙏

src/main/kotlin/ecommerce/interceptor/JwtAuthenticationInterceptor.kt
```diff
@@ -0,0 +1,26 @@
...
+    override fun preHandle(
+        request: HttpServletRequest,
+        response: HttpServletResponse,
+        handler: Any,
+    ): Boolean {
+        val accessToken =
+            request.getHeader("Authorization") ?: throw AuthorizationException("Authorization header missing")
+        if (!jwtProvider.validateToken(accessToken)) {
+            throw AuthorizationException("Invalid or expired JWT token")
+        }
```
Hyeon9mak: or, jwtProvider throw exception in itself! 😉 
```suggestion
        jwtProvider.validateToken(accessToken)
```

src/main/kotlin/ecommerce/controller/AdminController.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/admin")
+class AdminController(
+    private val cartService: CartService,
+) {
+    @GetMapping("/top-products")
+    fun findTop5ProductsInLast30Days(): ResponseEntity<List<TopProductStatResponse>> {
```
Hyeon9mak, line 17: 💯

src/main/kotlin/ecommerce/dto/ProductRequest.kt
Hyeon9mak: 💯 Perfect.
With this approach, fellow developers only need to look into the domain layer when the policy changes.

src/main/kotlin/ecommerce/handler/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,59 @@
...
+@ControllerAdvice
+class GlobalExceptionHandler {
+    private val logger = LoggerFactory.getLogger(GlobalExceptionHandler::class.java)
+
+    @ExceptionHandler(NoSuchElementException::class)
+    fun handleNoSuchElementException(e: NoSuchElementException): ResponseEntity<Void> {
+        logger.warn("Resource not found: ${e.message}", e)
+        return ResponseEntity.notFound().build()
+    }
```
Hyeon9mak, line 22: 😉 👍 

> When I see a 404, the first thought is usually “client asked for a resource that does not exist” which is exactly what handler returns.
I deliberately throw NoSuchElementException from the service layer when an ID is not found, so returning 404 is consistent. If I ever need to distinguish “wrong URL” from “entity missing”, it can be mapped to a different exception (e.g., MethodArgumentNotValidException → 400).So, I kept 404 for NoSuchElementException. It matches the “resource not found” scenario.
But please let me know your perspective about this. 🙏

There’s no definitive answer here, so Farhana’s perspective is also valid.
What’s important is the depth of thought and intentionality behind the decision. 😉 👍

src/main/kotlin/ecommerce/model/NewMember.kt
```diff
@@ -0,0 +1,8 @@
+package ecommerce.model
+
+data class NewMember(
+    val name: String,
+    val email: String,
+    val password: String,
+    val role: String,
+)
```
Hyeon9mak, line 8: `Member` and `NewMember`!
Separated two class for every use case! 👍 💯 
separate responsibility on use case one by one. 👍

src/main/kotlin/ecommerce/config/WebConfig.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@Configuration
+class WebConfig(
+    private val jwtAuthenticationInterceptor: JwtAuthenticationInterceptor,
+    private val loginMemberArgumentResolver: LoginMemberArgumentResolver,
+    private val adminRoleCheckInterceptor: AdminRoleCheckInterceptor,
+) : WebMvcConfigurer {
+    override fun addInterceptors(registry: InterceptorRegistry) {
+        registry.addInterceptor(jwtAuthenticationInterceptor)
+            .addPathPatterns("/api/protected/**")
```
Hyeon9mak, line 19: `protected` right? 🤔

Hyeon9mak (changes requested):
Hi Farhana!

U have reflected the requirements well and addressed the feedback nicely.
So I don't have many comments to add. 👍 

I'm leaving a 'Request changes' just in case you have any additional questions.
If not, feel free to submit the pull request again right away! 😉

src/main/kotlin/ecommerce/interceptor/JwtAuthenticationInterceptor.kt
farhanaahmed: You're right. It makes more sense. I have updated my code. Now, validateToken() throws AuthorizationException directly, and the redundant if blocks are gone. Thanks for the tip! 🙏

src/main/kotlin/ecommerce/config/WebConfig.kt
farhanaahmed: Thanks for your thoughtful review. I didn't notice this earlier. I have changed the inceptor path patterns and also updated all affected tests to reflect the new protected paths.

src/main/kotlin/ecommerce/infrastructure/JWTProvider.kt
```diff
@@ -8,13 +9,12 @@ import org.springframework.stereotype.Component
...
-    private lateinit var secretKey: String
-
-    @Value("\${security.jwt.token.expire-time-millis}") // <-- new key
-    private var validityInMillis: Long = 0
-
+    private val secretKey: String,
+    @Value("\${security.jwt.token.expire-time-millis}")
+    private val validityInMillis: Long,
+) {
     fun createToken(payload: String): String {
```
Hyeon9mak, line 18: nice

Hyeon9mak (approved):
Hi Farhana!
U did a great job progressing through this mission!
Hope the comments I left along the way were helpful. 🙏 

I’m approving this now
Wishing you lots of fun with the remaining missions! 😄 

Have a nice week ahead!

PR #30, opened 2025-07-28
[Spring-Ecommerce-Product Step 2] Submit knopgm's Mission

knopgm:
Hey!

I hope you are ok! 
This is my step2 of the spring-ecommerce mission.
First, I would like to mention that I got a cold in the middle of the project, which made me very slow at the beginning and a bit hurry at the end. 
I decided to Pull the project with somethings missing not to delay the review anymore, but here is somethings I will be doing while you review my code:

- Test ( at some point I gave up building with TDD or even building a test file, because all connections I have to create were bringing up new refactoring on things I already built, and tests were breaking all the time. So I decide to prioritaze the flow of code, test requests in postman and left the tests to do later). The tests I'm sending now is the ones I did before the complications start to arise.
- Interceptor interface. I still trying to understand it's necessity since all other stuff is working great without it. 

### Edge cases and plus (if time allows):
- Implement logic to deal with the case when a ADMIN user decided to delete a product that is already in the cart of clients. Should I delete it anyway from all places in the application? Or should I block the request? 🤔 
- UI. I'm not going to do this one, unless I see I have time to it, but I would like very much to improve it from what I have now.

Looking forward to you feedback! 🙏


src/main/kotlin/ecommerce/auth/JwtProvider.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import io.jsonwebtoken.Jwts
+import io.jsonwebtoken.SignatureAlgorithm
+import io.jsonwebtoken.security.Keys
+import java.util.*
+import javax.crypto.SecretKey
+
+object JwtProvider {
+    private val secretKey: SecretKey = Keys.secretKeyFor(SignatureAlgorithm.HS256)
+    private val EXPIRATION_TIME_MS = 3600_000 // 1 hour
```
Hyeon9mak: name with time unit and kind comment! 💯

src/main/kotlin/ecommerce/auth/JwtProvider.kt
```diff
@@ -0,0 +1,45 @@
...
+        return try {
+            val claims = Jwts.parserBuilder()
+                .setSigningKey(secretKey)
+                .build()
+                .parseClaimsJws(token)
+            !claims.body.expiration.before(Date())
+        } catch (e: Exception) {
+            false
+        }
+    }
```
Hyeon9mak:
```suggestion
    fun validateToken(token: String): Boolean {
        return try {
            val now = Date()
            val claims = Jwts.parserBuilder()
                .setSigningKey(secretKey)
                .build()
                .parseClaimsJws(token)
            claims.body.expiration.after(now)
        } catch (e: Exception) {
            false
        }
    }
```

`!` operator always make hard to read. 🤔

src/main/kotlin/ecommerce/service/AuthService.kt
```diff
@@ -0,0 +1,28 @@
...
+        }
+        return JwtProvider.generateToken(member.email)
+    }
+
+    fun extractAndValidateToken(authHeader: String): String {
+        val token = authHeader.removePrefix("Bearer ").trim()
+        if (!JwtProvider.validateToken(token)) {
+            throw IllegalArgumentException("Invalid token")
+        }
+        return JwtProvider.getEmailFromToken(token)
```
Hyeon9mak, line 27: JwtProvider can throw exception it self!

```kotlin
    fun validateToken(token: String): Boolean {
            val now = Date()
            val claims = Jwts.parserBuilder()
                .setSigningKey(secretKey)
                .build()
                .parseClaimsJws(token)
            require(claims.body.expiration.before(now)) { " ... " }
    }
```

I think it makes more readable & testable code!

src/main/kotlin/ecommerce/controller/AuthController.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @GetMapping("/members")
+    fun secure(@RequestHeader("Authorization") authHeader: String): ResponseEntity<String> {
+        return try {
+            val email = authService.extractAndValidateToken(authHeader)
+            ResponseEntity.ok("Hello, $email! Here are your products...")
+        } catch (e: IllegalArgumentException) {
+            ResponseEntity.status(401).body("Invalid token. $e")
+        }
+    }
```
Hyeon9mak: How about creating a separate ExceptionHandling class?
That way, your API documentation and tests can stay focused on the successful behavior of your endpoints!
https://docs.spring.io/spring-framework/reference/web/webmvc/mvc-controller/ann-advice.html#page-title

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,44 @@
...
+class CartController(private val cartService: CartService) {
+
+    @PostMapping("add/{productId}")
+    fun addToCart(
+        @LoginMember member: Member,
+        @PathVariable("productId") productId: Long
+    ): ResponseEntity<CartItem> {
+        val saved = cartService.addProductToCart(member.id, productId, 1)
+        return ResponseEntity.ok(saved)
+    }
```
Hyeon9mak: You might find it helpful to check the following documentation on how to represent ID paths!
https://restfulapi.net/resource-naming/

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    @DeleteMapping("/{productId}")
+    fun deleteCartItem(
+        @LoginMember member: Member,
+        @PathVariable productId: Long
+    ): ResponseEntity<Void> {
+        val deleted = cartService.removeItemFromCart(member.id, productId)
+        return if (deleted != null) ResponseEntity.noContent().build()
+        else ResponseEntity.notFound().build()
+    }
```
Hyeon9mak: Shall we try to recall something together?
When you encountered a 404 NOT FOUND error page, what was the first thing that came to your mind, Gabi?
Was it, 'I guess the data doesn’t exist,' or 'Did I type the wrong URL?'
Depending on that, we might need to reconsider which error code to return. 😄

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+
+    @PostMapping
+    fun create(
+        @RequestBody product: Product, @LoginMember member: Member
+    ): ResponseEntity<Product> {
+        if(member.role != Role.ADMIN) {
+            return ResponseEntity.status(401).build()
+        }
+        val savedProduct = productService.createProduct(product)
+        return ResponseEntity.created(URI.create("/products/${savedProduct.id}")).body(savedProduct)
```
Hyeon9mak: > Interceptor interface. I still trying to understand it's necessity since all other stuff is working great without it.

Interceptors come into play after the request passes through the filter and the DispatcherServlet, once the API path is determined.
In other words, they’re perfect for handling access control that varies per API endpoint — and it looks like you’ve found just the right use case for them. 😄

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    @GetMapping
+    fun read(): ResponseEntity<List<Product>> {
+        val products = productService.getAllProducts()
+        return if (products.isEmpty()) ResponseEntity.noContent().build()
+        else ResponseEntity.ok(products)
+    }
```
Hyeon9mak: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/204

check this! 😉

src/main/kotlin/ecommerce/dto/ProdutDTO.kt
```diff
@@ -0,0 +1,16 @@
...
+    val price: BigDecimal?,
+    val imageUrl: String?
+) {
+    fun validate(): ProductDTO {
+        require(!name.isNullOrBlank()) { "Name must not be blank." }
+        require(price != null && price > BigDecimal.ZERO) { "Price must be greater than 0." }
+        require(!imageUrl.isNullOrBlank() && imageUrl.startsWith("http")) { "Invalid image URL." }
+        return this
+    }
+}
```
Hyeon9mak: Request/Response DTOs and Domain Objects are used in the Presentation Layer and Domain Layer, respectively.
In your opinion, which layer is more appropriate for handling validation?

src/main/kotlin/ecommerce/repository/BaseProductStore.kt
```diff
@@ -0,0 +1,16 @@
...
+interface BaseProductStore {
+
+    fun isEmptyOrNull(): Boolean
+    fun count(): Int?
+    fun findAll(): List<Product>
+    fun insert(product: Product): Product
+    fun deleteById(id: Long): Int?
+    fun updateById(id: Long, product: Product): Int?
+    fun findByName(name: String): Product?
+    operator fun get(id: Long): Product? // products[1] // fun findById(id: Long): Product? // products.findById(1)
```
Hyeon9mak: Could u explain what this comment is referring to?!

src/main/kotlin/ecommerce/service/AuthService.kt
Hyeon9mak: and i think this exception more fit 😄 
`UnauthorizedException`

Hyeon9mak (changes requested):
Hi Gabi! nice work in step 2!

> Implement logic to deal with the case when a ADMIN user decided to delete a product that is already in the cart of clients. Should I delete it anyway from all places in the application? Or should I block the request? 🤔

As u know, It’s ultimately a matter of policy.
But if you have some time to spare, it might be more interesting to implement it so that deletion is not allowed.
If not, allowing all items to be deleted and spending more time learning Spring could also be a smart strategy. 😄 

> UI. I'm not going to do this one, unless I see I have time to it, but I would like very much to improve it from what I have now.

This isn’t a key requirement for the mission, but if you feel like trying it out, there’s no reason not to! 💪 

I left some comments. Feel free to ask any question!

src/main/kotlin/ecommerce/repository/BaseProductStore.kt
knopgm: I don't remember why this comment is there. I think my pair programmer and I was thinking about the function aloud and forgot to remove it after. 😅

src/main/kotlin/ecommerce/dto/ProdutDTO.kt
knopgm: Thank you for raising this question! I'm also a bit confused about so many layers and their many validations., even after understanding their responsibilities. 😅 
But for DTOs I found that is prefered to have inputs validated ealier then later.
What do you think about this validations? are them repetitions or ensuring safety?

src/main/kotlin/ecommerce/controller/ProductController.kt
knopgm: Interesting! GETs shouldn't return 204, but a empty list! 😮 
Refactored!

src/main/kotlin/ecommerce/controller/AuthController.kt
knopgm: Done!

src/main/kotlin/ecommerce/controller/ProductController.kt
knopgm: Now it is clear what a Interceptor is! Thank you! 😄 🙏

src/main/kotlin/ecommerce/controller/CartController.kt
knopgm: I implemented a AddToCartRequst and a DeleteCartItemRequest dtos to add the ID as a body. This way the path can be more clear with names like "/delete" and "/add".
Is that a good solution?

src/main/kotlin/ecommerce/controller/CartController.kt
knopgm: I updated the deletion request to this code:
```
@Protected
    @DeleteMapping("/delete")
    fun deleteCartItem(
        @RequestBody body: DeleteCartItemRequest,
        @LoginMember member: Member,
    ): ResponseEntity<Void> {
        val deleted = cartService.removeItemFromCart(member.id, body.productId)
        return if (deleted != null) {
            ResponseEntity.noContent().build()
        } else {
            ResponseEntity.notFound().build()
        }
    }
```
I think like this I can cover a 204 for the intention of the delete and a 404 if it was a bad URL.
Is that a good solution?

knopgm:
Hi GuMak!
Thank you for your quick and deep review of my code! I'm really enjoying to have someone that is deeply looking into the code and pushing me to reflect on best practices, decisions and subjects! xD
For this review, I implemented the step 2.4, improve the range of tests and carefully apply your review on top of what I mentioned.

The plus edge cases and improvements at the frontend, I'll have to leave for a future time.

Thank you again for your patience with my junior code mistakes and for your time! 🙏

src/main/resources/static/js/ProductManager.js
```diff
@@ -14,11 +14,11 @@ function productManager() {
...
             };
 
-            const res = await fetch(API_ENDPOINT, { ...options, method: "POST" });
+            const res = await fetch(`${API_ENDPOINT}/add`, { ...options, method: "POST" });
             return res;
         },
         async delete(id) {
-            const res = await fetch(`${API_ENDPOINT}/${id}`, { method: "DELETE" });
+            const res = await fetch(`${API_ENDPOINT}/delete`, { method: "DELETE" });
             return res;
```
Hyeon9mak, line 22: > I implemented a AddToCartRequst and a DeleteCartItemRequest dtos to add the ID as a body. This way the path can be more clear with names like "/delete" and "/add".
Is that a good solution?

HTTP methods inherently convey meaning—POST implies creation, and DELETE implies removal.
While it's important to express intent clearly, verbs like 'add', 'create', 'delete', or 'remove' are often redundant and can be omitted.

src/main/kotlin/ecommerce/auth/JwtProvider.kt
```diff
@@ -0,0 +1,50 @@
...
+import io.jsonwebtoken.Jwts
+import io.jsonwebtoken.SignatureAlgorithm
+import io.jsonwebtoken.security.Keys
+import java.util.Date
+import javax.crypto.SecretKey
+
+object JwtProvider {
+    private val secretKey: SecretKey = Keys.secretKeyFor(SignatureAlgorithm.HS256)
+    private const val EXPIRATION_TIME_MS = 3600_000 // 1 hour
+
```
Hyeon9mak:
```suggestion
    private val secretKey: SecretKey = Keys.secretKeyFor(SignatureAlgorithm.HS256)
    private const val EXPIRATION_TIME_MS = 3_600_000 // 1 hour
```
😉

src/main/kotlin/ecommerce/config/ExceptionHandler.kt
```diff
@@ -0,0 +1,19 @@
...
+    @ExceptionHandler(IllegalArgumentException::class)
+    fun handleIllegalArgument(ex: IllegalArgumentException): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.BAD_REQUEST).body("")
+    }
+
+    @ExceptionHandler(UnauthorizedException::class)
+    fun handleIllegalArgument(ex: UnauthorizedException): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body("")
+    }
+}
```
Hyeon9mak: What would happen if an exception occurs that we didn't explicitly handle?
And how can we prepare for such cases?
Hint: inheritance exception.

src/main/kotlin/ecommerce/controller/CartController.kt
```diff
@@ -0,0 +1,80 @@
...
+    @GetMapping("/active-members")
+    fun getActiveMembers(
+        @LoginMember member: Member,
+    ): ResponseEntity<List<MemberResponse>> {
+        if (member.role != Role.ADMIN) {
+            return ResponseEntity.status(401).build()
+        }
+        val items = cartService.getRecentActiveMembersInTheLast7Days()
+        if (items.isEmpty()) return ResponseEntity.noContent().build()
+        return ResponseEntity.ok(items)
```
Hyeon9mak: https://github.com/herotechcourse/spring-ecommerce-product/pull/30#discussion_r2249971778 👀 
need to refactor all api!

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -0,0 +1,69 @@
...
+    @DeleteMapping()
+    @Protected
+    @AdminOnly
+    fun delete(
+        @RequestBody body: DeleteProductRequest,
+        @LoginMember member: Member,
+    ): ResponseEntity<Void> {
+        if (member.role != Role.ADMIN) {
+            return ResponseEntity.status(401).build()
+        }
```
Hyeon9mak: what's different between status code 401 & 403?
and i think it can be handle in interceptor 🤔

src/main/kotlin/ecommerce/model/Member.kt
```diff
@@ -0,0 +1,22 @@
...
+    var role: Role,
+    var name: String,
+) {
+    init {
+        require(email.isNotBlank()) { "Email cannot be blank" }
+        require("@" in email && ".com" in email) { "Email should contain a @ and a .com" }
+        require(password.isNotBlank()) { "Password cannot be blank" }
+        require(password.length >= 4) { "Password with a minimum of 4  and maximum of 8 characters long" }
+        require(password.length <= 8) { "Password with a minimum of 4  and maximum of 8 characters long" }
+    }
```
Hyeon9mak, line 21: Looks like very good to generate test code 👀 
Haven't already forgot what u learned in past missions, have u?

Hyeon9mak (changes requested):
Hi Gabi, nice work to reflect feedback comments.

I Guess, it might conflict with the next mission.
So I only left comments on parts that could be improved without too much pressure.

Feel free to think it over and try things out at your own pace.
Let me know if you have any questions!

src/main/resources/static/js/ProductManager.js
knopgm: I realized that after. 🙈 😅

src/main/kotlin/ecommerce/config/ExceptionHandler.kt
knopgm: I implemented a `@ExceptionHandler` that inherit a `Exception` class to handle what is not falling in my 2 first exceptions: `IllegalArgumentException` and `UnauthorizedException`.
Thank you for catching that!  :)

src/main/kotlin/ecommerce/controller/ProductController.kt
knopgm:
```
401: means unauthorized and indicates that a request was not successful because it lacks valid authentication credentials for the requested resource.
```
```
403: is forbidden and indicates that the server understood the request but refused to process it because the request failure is tied to application logic, such as insufficient permissions to a resource or action.
```
I added a forbidden Exception that is called in the ExceptionHandler. I also fix my interceptor and the product controller test to correctly handle this kind of exception.

src/main/kotlin/ecommerce/controller/CartController.kt
knopgm: By reviewing your new comments and serching for contents, I decided that the path created for the use of id's in requests were and I rolled back my code to what it was before (except for those who I added a path just to name it, which was the case for /add and /delete.
For that reason, I think this comment can be marked as resolved.
Of course, as my reviewer, feel free to reopen this or anyother one you think is necessary. 😄

src/main/kotlin/ecommerce/controller/CartController.kt
knopgm: Thank you for catching this one. Turns out that I have to go back to study http response status code convention.
I did a general check in all Controllers and refactor their tests too. 👍

src/main/kotlin/ecommerce/model/Member.kt
knopgm: MemberTest implemented! 👍

knopgm:
Hi GuMak!
I'm sorry if it felt like I forget some past content.
As a beginner I'm getting used to bigger projects. I want to improve some other things for this project, but I'm also starting with the new content of the course.
Nevertheless, I'm open to refactor/implement all things you see that I need to better understand or complete this mission! 😎 
Thank you again for your time and patience with my juniority! 🙏

src/main/kotlin/ecommerce/config/ExceptionHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+    @ExceptionHandler(ForbiddenException::class)
+    fun handleIllegalArgument(ex: ForbiddenException): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.FORBIDDEN).body("${ex.message}")
+    }
+
+    @ExceptionHandler(Exception::class)
+    fun handleGeneralException(ex: Exception): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR)
+            .body("An unexpected error occurred: ${ex.message}")
+    }
```
Hyeon9mak, line 35: nice 👍

src/main/kotlin/ecommerce/controller/MemberController.kt
```diff
@@ -0,0 +1,40 @@
...
+    @GetMapping("/{memberId}")
+    fun get(
+        @PathVariable memberId: Long,
+        @LoginMember loggedMember: Member,
+    ): ResponseEntity<Member> {
+        val member = memberService.validateId(memberId)
+        if (member.id != loggedMember.id) {
+            return ResponseEntity.notFound().build()
+        }
+        return ResponseEntity.ok(member)
```
Hyeon9mak, line 38: Controller layer represent HTTP protocol on Kotlin code.
So compare member's id is not controller's responsibility.
What do u think about this? 😉

src/test/kotlin/ecommerce/model/MemberTest.kt
```diff
@@ -0,0 +1,73 @@
...
+                id = 1L,
+                email = "",
+                password = "1234",
+                role = Role.USER,
+                name = "Jon",
+            )
+        }.isInstanceOf(IllegalArgumentException::class.java)
+            .hasMessageContaining("Email cannot be blank")
+    }
+
```
Hyeon9mak, line 20: 😉 👍

Hyeon9mak (approved):
Hi Gabi,
Great job quickly applying the feedback!
glad to see you didn't miss improving the test code.

Since you're likely already doing a lot of refactoring in the spring-ecommerce-order mission,
I just left a few final, simple comments.
It’d be great if you could consider them and make some small improvements in the order mission.

Hope you enjoy the rest of the mission!

PR #31, opened 2025-07-28
[spring-ecommerce-product - Step 2] Submit san-ghun's mission

san-ghun:
# Description

This PR is an implementation of a simple HTTP API that allows users to retrieve, add, update, and delete products and manipulation with dedicated cart for user, using Spring Boot framework with Kotlin, also using H2 in-memory database for development purpose. 
With step 2 requirements, the implementation provide validation for request, handling exceptions, user authentication with registration and login, user authorization, cart api, and cart statistics api.

# Changes

- Introduce `Member` and `CartItem` models to handle the authentication and cart API
- Introduce `AuthController`, `CartController`, `CartStatsController` to provide endpoints to access Auth and Cart API
- Introduce `AuthService` and `CartService` to handle logic for Auth and Cart API
- Introduce `GlobalExceptionHandler` to handle exceptions
- Introduce `JwtTokenProvider` to serve authentication of member, using JWT
- Introduce `LoginMemberArgumentResolver` to automatically authenticate member
- Introduce `CheckAdminInterceptor` to filter access for admin specific endpoints by checking member's role
- Create table `member` to store member data
- Create table `cart_item` to store cart data
- Create table `cart_item_event` to store event data for cart to provide data for cart statistics

# Testable Endpoints

- `POST /api/products` - create a product
- `GET /api/products` - read all products
- `GET /api/products/{id}` - read a product with id
- `PUT /api/products/{id}` - update a product with id
- `DELETE /api/products/{id}` - delete a product with id
- `GET /` - provide view of list of all products
- `GET /add/product` - provide view of form to add new product
- `GET /update/product/{id}` - provide view of form to update a product

- `POST /api/members/register` - register a member
- `POST /api/members/login` - login a member
- `GET /api/members/me` - get member data

- `GET /api/cart` - view user cart items
- `POST /api/cart` - add item to user cart
- `DELETE /api/cart/{productId}` - remove an item from user cart
- `PUT /api/cart/{productId}` - update quantity of an item in user cart

- `GET /api/admin/cart-stats/top5-products` - get top 5 added products inn last 30 days
- `GET /api/admin/cart-stats/active-members` - get active members in last 7 days

# Additional Note

I was a bit in hurry to meet the time limit for submission. I will look forward to you review! 

Thank you!


build.gradle.kts
```diff
@@ -22,12 +22,20 @@ repositories {
...
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
+    implementation("org.springframework.boot:spring-boot-starter-validation")
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
+    implementation("io.jsonwebtoken:jjwt-api:0.12.6")
+
     runtimeOnly("com.h2database:h2")
+    runtimeOnly("io.jsonwebtoken:jjwt-impl:0.12.6")
+    runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.12.6")
```
pkeugine, line 33: What are the differences between `implementation` and `runtimeOnly`?
I'm curious about the reason why you used `implementation` for `jjwt-api` and `runtimeOnly` for others.

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+        } catch (e: IllegalArgumentException) {
+            println(e.message)
+            false
+        }
+    }
+
+    companion object {
+        private const val SECRET_KEY = "X1yk3dV+NQemyyJbg6WzioaRR1NArxZxjMvvroTZauk="
+        private const val EXPIRATION_TIME = 3600000L
+    }
```
pkeugine: I recommend moving this information into `application.properties` file and manage these there instead of keeping as part of application logic. One of the reason is because properties that could change over time are better kept separately, and other reason is because we want to hide it from public, even version control systems. (especially `SECRET_KEY`)

Keeping them in `application.properties` won't really hide them since it's already in our git history, but I think this would be a great time for us to find out how we can hide them. We don't have to apply it to this project, but we'll be able to apply it to the next project when we use a secret key!

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+            false
+        } catch (e: IllegalArgumentException) {
+            println(e.message)
+            false
+        }
+    }
+
+    companion object {
+        private const val SECRET_KEY = "X1yk3dV+NQemyyJbg6WzioaRR1NArxZxjMvvroTZauk="
+        private const val EXPIRATION_TIME = 3600000L
```
pkeugine: Also, what's the unit of `3600000L`? Seconds? Milliseconds?
It would be great for us to show it somewhere so we don't make a mistake later on.

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,57 @@
...
+        return ResponseEntity.badRequest().body(errorBody)
+    }
+
+    @ExceptionHandler(AuthorizationException::class)
+    fun handleAuthorizationException(e: AuthorizationException): ResponseEntity<Map<String, Any>> {
+        val error = mapOf("authorization" to e.message)
+        val errorBody = mapOf("errors" to error)
+        println("AuthorizationException occurred: $errorBody")
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(errorBody)
+    }
```
pkeugine: When it comes to response body of `UNAUTHORIZED` response, I prefer not giving it a body with error information, unless we really need it.
This is because we don't want people to find out what's wrong when they fail to breach into our system. I'm curious about your opinion on this :)

README.md
```diff
@@ -1,88 +1,410 @@
...
+- [x] override `resolveArgument()` to handle access token and return the member
+
+#### LoginMember
+
+- [x] is an annotation class for authorization of member, using argument resolver 
+
+#### CheckAdminInterceptor
+
+- [x] has private property as input `AuthService`
+- [x] override `preHandle()` to intercept and process admin member filtering
```
pkeugine, line 299: Wow, nice work updating README 😮 👍

src/main/kotlin/ecommerce/auth/AuthorizationExtractor.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import jakarta.servlet.http.HttpServletRequest
+
+interface AuthorizationExtractor<T> {
+    fun extract(request: HttpServletRequest): T
+
+    companion object {
+        const val AUTHORIZATION = "Authorization"
+    }
+}
```
pkeugine, line 11: Since there is currently a single implementation of this interface, I'm curious about the reason why you chose to keep this as an interface! It's not wrong, but some developers like me actually prefer to make an interface when we actually need it.

src/main/kotlin/ecommerce/auth/BearerAuthorizationExtractor.kt
```diff
@@ -0,0 +1,28 @@
...
+            if (value.lowercase().startsWith(BEARER_TYPE.lowercase())) {
+                var authHeaderValue = value.substring(BEARER_TYPE.length).trim()
+                request.setAttribute(ACCESS_TOKEN_TYPE, value.substring(0, BEARER_TYPE.length).trim())
+                val commaIndex = authHeaderValue.indexOf(',')
+                if (commaIndex > 0) {
+                    authHeaderValue = authHeaderValue.substring(0, commaIndex)
+                }
+                return authHeaderValue
+            }
+        }
```
pkeugine:
```suggestion
        while (headers.hasMoreElements()) {
            val value = headers.nextElement()
            if (value.lowercase().startsWith(BEARER_TYPE.lowercase())) {
                val authHeaderValue = value.substring(BEARER_TYPE.length).trim()
                request.setAttribute(ACCESS_TOKEN_TYPE, value.substring(0, BEARER_TYPE.length).trim())
                val commaIndex = authHeaderValue.indexOf(',')
                if (commaIndex > 0) {
                    return authHeaderValue.substring(0, commaIndex)
                }
                return authHeaderValue
            }
        }
```

We can prevent `var` by returning the value immediately :)

src/main/kotlin/ecommerce/controller/api/AuthController.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.net.URI
+
+@RestController
+@RequestMapping("/api/members")
+class AuthController(
+    private val authService: AuthService,
+) {
+    private val authorizationExtractor: AuthorizationExtractor<String> = BearerAuthorizationExtractor()
```
pkeugine: Could there be a way for us to inject `authorizationExtractor` in constructor just like `authService`? 🤔

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,292 @@
...
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
+class AuthControllerTest {
+    @Autowired private lateinit var jdbcTemplate: JdbcTemplate
+
+    @Autowired private lateinit var jdbcMemberDAO: JdbcMemberDAO
+
+    @Autowired private lateinit var jwtTokenProvider: JwtTokenProvider
+    private lateinit var authService: AuthService
+    private lateinit var controller: AuthController
```
pkeugine:
```suggestion
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
class AuthControllerTest {
    @Autowired private lateinit var jdbcTemplate: JdbcTemplate

    @Autowired private lateinit var jdbcMemberDAO: JdbcMemberDAO

    @Autowired private lateinit var jwtTokenProvider: JwtTokenProvider
    @Autowired private lateinit var authService: AuthService
    @Autowired private lateinit var controller: AuthController
```
Since we're using `@SpringBootTest`, why don't we inject services and controllers as well?

To go further, if we have test coverage for the controller using `RestAssured`, do we need another test for `AuthController`? This is not a right or wrong answer, so feel free to tell me your personal opinion :)

In my point of view, `registerMember()` test is a down graded version of `registerMember() - should insert member and return 201 when form is valid`. We could have both, but as the project gets bigger and more tests get created, I think having both for all controllers will result in more time spent creating and managing them.

Still, I want to point out that you did very well on creating fine-grained tests 👍

pkeugine: This also leads to another question : `Which tests do you think are necessary when maintaining projects?`

src/main/resources/schema.sql
```diff
@@ -1,7 +1,49 @@
...
+    member_id  LONG NOT NULL,
+    product_id LONG NOT NULL,
+    created_at TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (id),
+
+    FOREIGN KEY (member_id) REFERENCES member (id),
+    FOREIGN KEY (product_id) REFERENCES product (id)
+);
+
+// check for how to creat index!
```
pkeugine: Let's remove things like `-- DROP TABLE cart_item CASCADE;` and `// check for how to creat index!` before we merge this PR into the main branch :)

Besides, `//` is not a valid syntax anyways.

src/main/resources/schema.sql
```diff
@@ -1,7 +1,49 @@
...
-                         imageUrl    TEXT    NOT NULL,
-                         PRIMARY KEY (id)
+-- DROP TABLE product CASCADE;
+
+CREATE TABLE product
+(
+    id       LONG         NOT NULL AUTO_INCREMENT,
+    name     VARCHAR(255) NOT NULL,
+    price    DOUBLE       NOT NULL,
+    imageUrl VARCHAR(255) NOT NULL,
```
pkeugine: I'm curious about the types you've used when creating table : `LONG`, `VARCHAR`, and `DOUBLE`.
What is the difference between `LONG` and `Long` in Kotlin?
There's also `BIGINT`, so it would be nice to check that out too :)

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,292 @@
...
+                PRIMARY KEY (id)
+            );""",
+        )
+
+        val query =
+            """INSERT INTO member (email, password, role) VALUES ( 'san@htc.com', 'san1234', 'admin');
+            INSERT INTO member (email, password, role) VALUES ( 'dan@htc.com', 'dan1234', 'admin');
+            INSERT INTO member (email, password) VALUES ( 'ann@htc.com', 'ann1234');
+            INSERT INTO member (email, password) VALUES ( 'min@htc.com', 'min1234');"""
+        jdbcTemplate.batchUpdate(query)
```
pkeugine: It would be great to find out other ways of adding test fixtures before each tests, and compare them.
After that, let's see if there are any better options we could choose for our situation.
(It's also fine to keep as is, but I want to know what you think!)

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,364 @@
...
+        val targets =
+            listOf(
+                "Contains unallowed character",
+                "Product name is required",
+                "Must be no more than 15 characters, including spaces",
+            )
+        val resBody = response.jsonPath().getMap<String, String>("errors")
+        val actual = resBody["name"]
+        Assertions.assertThat(response.statusCode()).isEqualTo(HttpStatus.BAD_REQUEST.value())
+        Assertions.assertThat(actual).isIn(targets)
```
pkeugine: You've made a lot of tests 👍 
In certain test classes, `assertThat` is used right away while in others, it's called with `Assertions` before it.
Why don't we keep it the same?

This might seem minor, but I like to keep things consistent as much as I can, since we never know where things might go wrong when we don't keep the consistency.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,364 @@
...
+                )
+                .contentType(ContentType.JSON)
+                .`when`().post("/api/products")
+                .then().log().all().extract()
+
+        val expected = "Must be no more than 15 characters, including spaces"
+        val resBody = response.jsonPath().getMap<String, String>("errors")
+        val actual = resBody["name"]
+        Assertions.assertThat(response.statusCode()).isEqualTo(HttpStatus.BAD_REQUEST.value())
+        Assertions.assertThat(actual).isEqualTo(expected)
```
pkeugine:
```suggestion
        RestAssured
            .given().log().all()
            .body(
                Product(
                    name = "this is very long name",
                    price = 1.5,
                    imageUrl = "https://www.product.com/image/1",
                ),
            )
            .contentType(ContentType.JSON)
            .`when`().post("/api/products")
            .then().log().all()
            .assertThat()
            .statusCode(HttpStatus.BAD_REQUEST.value())
            .body("errors.name", equalTo("Must be no more than 15 characters, including spaces"))
```

This is not a must, but we can also let `RestAssured` assert everything :)

src/main/kotlin/ecommerce/controller/api/CartController.kt
```diff
@@ -0,0 +1,80 @@
...
+        @PathVariable productId: Long,
+        @LoginMember member: Member,
+    ): ResponseEntity<String> {
+        val memberId = member.id ?: throw InternalServerErrorException("auth failed")
+        val affectedRows = cartService.removeFromCart(memberId, productId)
+        when (affectedRows) {
+            1 -> return ResponseEntity.ok().body(MESSAGE_REMOVE_SUCCESS)
+            0 -> throw NotFoundException("Product not found in the cart - Product ID: $productId")
+            else -> throw InternalServerErrorException("Unexpected delete on item in cart - Product ID: $productId")
+        }
```
pkeugine: Instead of having this logic in controller, why don't we let service validate if the product exists before removing it?
It might seem like we're querying DB one more time when we can just utilize the result, so I'm curious about your opinion on this.

pkeugine: To think further, someone might say : `If there isn't a product in the first place, shouldn't delete request be treated as succeeded request?`

What's your thought on this? 😄

src/main/kotlin/ecommerce/dao/JdbcCartDAO.kt
```diff
@@ -0,0 +1,156 @@
...
+import ecommerce.dto.TopProductStats
+import ecommerce.model.CartItem
+import org.springframework.jdbc.core.JdbcTemplate
+import org.springframework.jdbc.core.RowMapper
+import org.springframework.jdbc.support.GeneratedKeyHolder
+import org.springframework.stereotype.Repository
+import java.sql.ResultSet
+
+@Repository
+class JdbcCartDAO(private val db: JdbcTemplate) : CartDAO {
```
pkeugine: Usually, [class names with acronyms also follow the rule of camel case](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case).
So in this case, the class name would be `JdbcCartDao`.
Please apply this to other parts of the project as well.

src/main/kotlin/ecommerce/controller/api/CartController.kt
```diff
@@ -0,0 +1,80 @@
...
+            1 -> return ResponseEntity.ok().body(MESSAGE_UPDATE_SUCCESS)
+            0 -> throw NotFoundException("Product not found in the cart - Product ID: $productId")
+            else -> throw InternalServerErrorException("Unexpected update on item in cart - Product ID: $productId")
+        }
+    }
+
+    companion object {
+        const val MESSAGE_ADD_SUCCESS = "Item added to cart"
+        const val MESSAGE_REMOVE_SUCCESS = "Item removed from cart"
+        const val MESSAGE_UPDATE_SUCCESS = "Item updated in cart"
```
pkeugine: Since you've extracted the message into constants, why don't we do that to exception message as well? (Or the other way around).

To me, keeping these messages separately causes me to go back and forth this file to see what gets returned.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,364 @@
...
+        jdbcTemplate.execute("DROP TABLE product CASCADE")
+        jdbcTemplate.execute(
+            """CREATE TABLE product (
+                         id          LONG    NOT NULL AUTO_INCREMENT,
+                         name        varchar(255)    NOT NULL,
+                         price       DOUBLE  NOT NULL,
+                         imageUrl    TEXT    NOT NULL,
+                         PRIMARY KEY (id)
+                    );""",
+        )
```
pkeugine: I didn't ask this in the last step, but it's interesting to see that you're `DROP`ping the table and creating it before each tests. Is there a reason for this? 😉

pkeugine (changes requested):
Hello San, and well done updating the project 👍 
Your tests look solid, and application logic was very clean for me to follow.

After adding some reviews, I realized that there are already like 20 of them...
so I'm requesting a change before we talk more.

There are questions that I thought was interesting for us to talk about at around this time of mission, and most of them don't have one perfect answer so feel free to answer them with ease.

If you have any questions, leave a comment or send me a DM! Request another review once you're done.

Happy Hacking! 😄 😄 😄

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
san-ghun: Yes, it would be better to get secret value not from source code, but from external source. 
I will move the parts from the file to `application.properties`. And, I will figure out how to do that on next project!

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
san-ghun: Yes, thanks for pointing out. I totally missed the detail. I will keep that in mind!

build.gradle.kts
san-ghun: `implementation` is for using the dependency on compilation. `runtimeOnly` is for using the dependency on runtime. 
If I put `jjwt-api` in `runtimeOnly` the program will fail on compilation, which means I can not run program including test in IDE.
If I put other `runtimeOnly` dependencies into `implementation`, it will compile but having bigger size and maybe include unnecessary API.

san-ghun: Haven't thought it seriously before. Thanks for pointing out and asked for me!

src/main/kotlin/ecommerce/GlobalExceptionHandler.kt
san-ghun: Oh, that makes sense. Thanks, I will remove body content for response!

README.md
san-ghun: Thank you!

src/main/kotlin/ecommerce/auth/AuthorizationExtractor.kt
san-ghun: I followed an article from web and learning material to implement the authorization extractor. The article gave me several options for implementations of authorization extractor, using the interface. And I only picked `BearerAuthorizationExtractor` and not others, that is the reason why the interface remained. 

What I've thought was, if considering the possibility of applying multiple authorization methods, it might be good to have a unified form of object for authorization extractor. 

However, I am not exactly sure is there any way to implement multiple auth method in project?

src/main/kotlin/ecommerce/auth/BearerAuthorizationExtractor.kt
san-ghun: Yeah! I love it. Thanks!!

src/main/kotlin/ecommerce/controller/api/AuthController.kt
san-ghun: I may add `@Component` annotation for `BearerAuthorizationExtractor` to let Spring to catch auth extractor as a bean...(?)

```kotlin
@Component
class BearerAuthorizationExtractor : AuthorizationExtractor<String> {
// ...
}

class AuthController(
    private val authService: AuthService,
    private val authorizationExtractor: AuthorizationExtractor<String> = BearerAuthorizationExtractor(),
) {
// ...
}
```

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
san-ghun: > ... we have test coverage for the controller using `RestAssured`, do we need another test for `AuthController`?

The reason was that, I actually not know well and not sure about `RestAssured`. I couldn't guarantee what if, how it work in different test environment depends on system setting. So I tried to have both test cases using `RestAssured` and just test library for my own safety.


> This also leads to another question : `Which tests do you think are necessary when maintaining projects?`

I think it depends on situation and size of an application. However, End-to-End test for an application is must, and other cases are needed to guarantee the functionality of logic, including inputs and outputs of objects and methods.

src/main/resources/schema.sql
san-ghun: Yes, thank you! I agree.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
san-ghun: Thank you! I forgot about that. i used auto formatting with `ktlint` and it automatically added notation like that. I fixed it.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
san-ghun: OHHHHH! So cool! I will to that!

src/main/kotlin/ecommerce/dao/JdbcCartDAO.kt
san-ghun: Okay, I will rename those.

src/main/kotlin/ecommerce/controller/api/CartController.kt
san-ghun: Yeah, it seems better with constant messages, both for success and exception.

src/main/kotlin/ecommerce/controller/api/CartController.kt
san-ghun: > Instead of having this logic in controller, why don't we let service validate if the product exists before removing it? It might seem like we're querying DB one more time when we can just utilize the result, so I'm curious about your opinion on this.

Yes, it totally make sense. There should be check for the existence of product before removing it from cart. Thanks for pointing out!


> To think further, someone might say : `If there isn't a product in the first place, shouldn't delete request be treated as succeeded request?`

Since it regards with user experience and user interaction, I think I have to discuss it with other related entities like, frontend engineer and product manager. 
For my opinion, I like to treat those differently because the result and action are different.

What do you think??

src/main/resources/schema.sql
san-ghun: Oh...! I didn't noticed the difference between `LONG` in SQL and `Long` in Kotlin.

SQL Type | Kotlin Type | Notes
-- | -- | --
BIGINT | Long | Standard mapping for large ints
INT, INTEGER | Int | 32-bit integer
LONG (Oracle) | String | Treated as large character/text
CLOB, TEXT | String | For large text data

Thanks!

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
san-ghun: Yes, I also thought there should be better way to replace those boilerplate like code.
So I am thinking to apply `@Sql()` annotation like this,

```kotlin
@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT)
@Sql(
    scripts = ["/sql/member.sql", "/sql/product.sql", "/sql/cart_item.sql", "/sql/cart_item_event.sql"],
    executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD,
)
class CartControllerTest(
    @Autowired private val controller: CartController,
) {
    // ...
}
```

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
san-ghun: I thought "drop"ping a table and create new one before each test method is easier to maintain the isolation of test, rather than deleting all records each time. That was the reason why I do like that. 

I've heard about using `TRUNCATE TABLE table_name;` , but the query didn't work in my test.

san-ghun:
Hey PK!

Hope you the best in this summer!
Always thanks for your review with valuable questions.

I tried to apply your reviews. And most of changes are done in test files.

## Main Changes

- Add `.sql` files in resource for test utility purpose to execute during test.
- Apply `@Sql()` annotation with SQL files to remove boilerplate code blocks used with `@BeforeEach` `fun setup()`.
- Apply `RestAssured` style assertions in controller tests.

## Additional Note

I really appreciate your thoughtful comments that make me learn and study even deeper.
Thank you!

### ref) Questions

- Difference between `implementation` and `runtimeOnly` in `build.gradle.kts` file?
	- What happen if dependency used in different place?
- Where should I store **secret** in project? 
	- `application.properties` or some where hidden?
- What are the differences between `LONG` in SQL and `Long` in Kotlin?
- When implementing `DELETE`, if there isn't a `product` in the first place, shouldn't delete request be treated as succeeded request? 
	- Should it be treated as `OK` or `Not Found` or `No Content` ?
- Difference between `DROP` and `TRUNCATE` in SQL?

src/main/kotlin/ecommerce/auth/AuthorizationExtractor.kt
pkeugine: I do agree that most of the times, a project doesn't require multiple options for extracting authorization related information. That's also why I think introducing interface could be done later on, when we actually need it.

If you ask `so when would we use other method?`, I would say
* when we're dealing with legacy system, especially ones that already made their own process of authorization
* when we're not making a web application server that deals things with HTTP
* or even when we're dealing with system that has some mistakes (this happens quite often, since people are not always perfect)

src/main/kotlin/ecommerce/controller/api/AuthController.kt
pkeugine: Nice! That would also mean that we could keep things like : 
```kotlin
class AuthController(
    private val authService: AuthService,
    private val authorizationExtractor: AuthorizationExtractor<String>,
) {
```

src/main/kotlin/ecommerce/controller/api/CartController.kt
pkeugine: Nice!
I would also do the same thing : talk with teammates.
If I had to choose myself, I would actually let it succeed if the resource is not there in the first place.
Remember [`idempotency`](https://github.com/herotechcourse/spring-ecommerce-product/pull/7#discussion_r2225732617)? 

[`DELETE` method's official characteristics contain it according to RFC](https://datatracker.ietf.org/doc/html/rfc9110#section-9.2.2) :)

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
pkeugine: Using what works for you seems alright 👍

pkeugine (approved):
Hello San, that refactoring was actually very quick and clean 👍 
You've done well answering to my questions, and I think we're good with this mission.
Since you're already on the next mission, I left some reactions and my thoughts on our conversation rather than requesting for more refactoring. (There doesn't seem much anyways 😮 )

I hope this mission helped you understand more about Spring Boot and web application server, and I'm sure you'll do well on your next one.

As always,
Happy Hacking! 🔥 🔥 🔥

PR #32, opened 2025-07-28
[spring-ecommerce-product - Step-2] Submit Apoorva's mission

Apoorva-S-Natesh:
**Pull Request: Core User Features (Authentication, Cart, Admin Statistics)**
This Pull Request introduces foundational user-facing and administrative functionalities to the e-commerce application, encompassing user authentication, shopping cart management, and basic administrative statistics.

Throughout this implementation, several challenges were encountered and resolved, providing valuable learning:

1. Optional vs. Nullable Types:
Struggle: Initial confusion about using java.util.Optional versus Kotlin's native nullable types (Type?).
Resolution: Refactored repositories and services to consistently use Kotlin's nullable types (Type?) for methods that might return no result, leading to more idiomatic Kotlin code.

2. UNCHECKED_CAST Warning in JwtTokenProvider:
Struggle: Compiler warning when casting Any? to List<String> for JWT claims.
Resolution: Used filterIsInstance<String>() to provide a type-safe way to filter and cast collection elements, eliminating the warning without suppression.

3. BadSqlGrammarException (LAST_INSERT_ID()):
Struggle: 500 Internal Server Error due to using MySQL-specific SELECT LAST_INSERT_ID() in CartItemRepository while using H2 database.
Resolution: Corrected the SQL to SELECT IDENTITY() for H2.

4. InvalidDataAccessApiUsageException (getKey() with multiple generated keys):
Struggle: After fixing BadSqlGrammarException, a new 500 Internal Server Error occurred because GeneratedKeyHolder.getKey() expected a single key but H2 returned multiple (ID and created_at).
Resolution: Modified jdbcTemplate.update in CartItemRepository.save() to explicitly specify arrayOf("id") as the column(s) for which to return generated keys, ensuring only the id is retrieved.

5. 500 Internal Server Error (Unhandled ProductNotFoundException in getCartItems):
Struggle: POST /api/cart/items returned 500 even when the item was added, because CartService.getCartItems (called by GET /api/cart/items) didn't handle ProductNotFoundException for potentially deleted products, causing it to fall through to the generic 500 handler.
Resolution: Implemented a try-catch with mapNotNull in CartService.getCartItems to gracefully handle ProductNotFoundException for individual cart items, logging warnings and filtering out invalid items from the response.

6. 403 Forbidden on Cart Endpoints (Initial LoginMemberArgumentResolver Integration):
Struggle: Initial attempts to use LoginMemberArgumentResolver resulted in 403 Forbidden errors on cart endpoints. This was due to the resolver not correctly fetching the Member object or not being properly registered.
Resolution: Ensured LoginMemberArgumentResolver correctly retrieves the Member from SecurityContextHolder and MemberService, and that WebConfig correctly registers it.

7. All Test run together fails due to spring security and Dirties Context classMode - Before Each method
Need to improve: I need to remove Dirties Context as the tests are very slow, unhealthy, and dirty! Should Implement @query and Truncate. 
Should also implement a AbstractintegratiionTest superclass which will have the test setup for all the Integration tests like registering, logging in and the cleaning up. which can then be inherited, and code repetition can be cut.

8. The unique constraint in the Schema caused issue when the tests had repeated names.
Resolution: Test was refactored to contain names that did not repeat

This mission was really tough as I am new to Spring and JWT, and I have tried implementing the basic requirements and decided to skip the optional HandlerInterceptor to check the user's role for now. I want to learn more about Spring and implement it later. I learnt a lot from this mission and truly value your feedback.

Thank you for your time and review!


README.md
```diff
@@ -23,3 +28,38 @@
...
+- [] Retrieve top 5 products added to any cart
+  - ranking based on count of additions
+  - tie breaking - most recent add listed first
+  - response includes - product name, total times added, most recent added time
+- [] Recently Active Members 
+  - List of members who added at least one product in last 7 days
+  - response includes - member ID, name, email
+- [] Role based access control(Admin only)
+  - statistical API endpoints - admins only
+  - 401 Unauthorized if user is not authorized
```
da-nyee: It looks like you’ve already implemented all the features — what about checking the boxes?

src/main/kotlin/ecommerce/controller/product/ProductController.kt
```diff
@@ -20,18 +21,18 @@ import java.net.URI
 class ProductController(private val productService: ProductService) {
```
da-nyee, line 23:
```suggestion
@RequestMapping("/products")
class ProductController(private val productService: ProductService) {
    // implementation

    @GetMapping("/{id}")
    fun getProductById(
        @PathVariable("id") id: Long,
    ): ResponseEntity<ProductResponse> {
        // implementation
    }
    
    @PutMapping("/{id}")
    fun updateProduct(
        @PathVariable("id") id: Long,
        @Valid @RequestBody request: UpdateProductRequest,
    ): ResponseEntity<ProductResponse> {
        // implementation
    }
}
```

You can use `@RequestMapping` like this, since all the APIs share the same `/products` prefix.

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,66 @@
...
+    @ResponseStatus(HttpStatus.NOT_FOUND)
+    fun handleProductNotFound(ex: ProductNotFoundException): Map<String, String> {
+        return mapOf("error" to ex.message.orEmpty())
+    }
+
+    @ExceptionHandler(DuplicateProductNameException::class)
+    @ResponseStatus(HttpStatus.CONFLICT)
+    fun handleDuplicateException(ex: DuplicateProductNameException): Map<String, String> {
+        return mapOf("error" to ex.message.orEmpty())
+    }
```
da-nyee:
```suggestion
    @ExceptionHandler(value = [
        DuplicateProductNameException::class,
        EmailAlreadyExistsException::class,
    ])
    @ResponseStatus(HttpStatus.CONFLICT)
    fun handleConflictException(ex: RuntimeException): Map<String, String> {
        return mapOf("error" to ex.message.orEmpty())
    }
```

You can group these exceptions together like this, as they return the same HTTP status.

src/main/kotlin/ecommerce/dto/product/UpdateProductRequest.kt
```diff
@@ -0,0 +1,28 @@
+package ecommerce.dto.product
+
+import jakarta.validation.constraints.Min
+import jakarta.validation.constraints.NotNull
+import jakarta.validation.constraints.Pattern
+import jakarta.validation.constraints.Size
+
+data class UpdateProductRequest(
```
da-nyee: `CreateProductRequest` and `UpdateProductRequest` seem very similar.
I think it'd be better to combine them into one 👀

src/main/kotlin/ecommerce/dto/product/UpdateProductRequest.kt
```diff
@@ -0,0 +1,28 @@
...
+    @field:NotNull(message = "Name must not be blank")
+    @field:Size(max = 15, message = "Name must be at most 15 characters")
+    @field:Pattern(
+        regexp = "^[a-zA-Z0-9 ()\\[\\]+\\-&/_]{1,15}$",
```
da-nyee: Regex 👍

src/main/kotlin/ecommerce/service/LoginMemberArgumentResolver.kt
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.security.core.userdetails.User
+import org.springframework.stereotype.Component
+import org.springframework.web.bind.support.WebDataBinderFactory
+import org.springframework.web.context.request.NativeWebRequest
+import org.springframework.web.method.support.HandlerMethodArgumentResolver
+import org.springframework.web.method.support.ModelAndViewContainer
+
+@Target(AnnotationTarget.VALUE_PARAMETER)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class LoginMember
```
da-nyee: Custom annotation 👍

da-nyee: Tiny suggestion — maybe move this to a separate file?

src/main/kotlin/ecommerce/service/LoginMemberArgumentResolver.kt
```diff
@@ -0,0 +1,56 @@
...
+
+@Component
+class LoginMemberArgumentResolver(
+    private val memberService: MemberService,
+) : HandlerMethodArgumentResolver {
```
da-nyee, line 20: Nice work — you're handling the authentication 👍

da-nyee: But it looks like there's no authorization check — how about adding it?

da-nyee: And I have one question — what's the difference between `Authentication` and `Authorization`?
I'd love to hear your take on it ☺️

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,123 @@
...
+            null
+        }
+    }
+
+    fun findAllByMemberId(memberId: UUID): List<CartItem> {
+        val sql = "SELECT * from cart_items WHERE member_id = ?"
+        return jdbcTemplate.query(sql, carItemRowMapper, memberId.toString())
+    }
+
+    fun findByProductId(productId: Long): CartItem? {
```
da-nyee: Seems like it's not used 👀

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,123 @@
...
+        val sql = "SELECT * from cart_items WHERE product_id = ?"
+        return jdbcTemplate.query(sql, carItemRowMapper, productId).firstOrNull()
+    }
+
+    fun deleteById(id: Long) {
+        val sql = "DELETE FROM cart_items WHERE id = ?"
+        jdbcTemplate.update(sql, id)
+    }
+
+    fun deleteByMemberIdAndProductId(
```
da-nyee: This one as well 👀

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,123 @@
...
+                cart_items ci
+            JOIN
+                products p ON ci.product_id = p.id
+            WHERE
+                ci.created_at >= ?
+            GROUP BY
+                p.id, p.name
+            ORDER BY
+                times_added DESC, most_recent_added_time DESC
+            LIMIT 5
```
da-nyee, line 89: You nailed it 💯

da-nyee: How would you add indexes to `cart_items` and `products`?

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,123 @@
...
+                mostRecentAddedTime = rs.getObject("most_recent_added_time", LocalDateTime::class.java),
+            )
+        }, since)
+    }
+
+    fun findRecentlyActiveMemberIds(since: LocalDateTime): List<UUID> {
+        val sql = """
+            SELECT DISTINCT ci.member_id
+            FROM cart_items ci
+            WHERE ci.created_at >= ?
```
da-nyee: I think this query needs to be fixed — how about using a `JOIN` to meet the requirement?

src/main/kotlin/ecommerce/security/JwtRequestFilter.kt
```diff
@@ -0,0 +1,29 @@
...
+import jakarta.servlet.FilterChain
+import jakarta.servlet.http.HttpServletRequest
+import jakarta.servlet.http.HttpServletResponse
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.stereotype.Component
+import org.springframework.web.filter.OncePerRequestFilter
+
+// Interacts with JwtTokenProvider and SpringSecurity's context, not with the database directly
+@Component
+class JwtRequestFilter(
```
da-nyee, line 12: You used a `Filter` 👍

da-nyee: It might be helpful to study how `DispatcherServlet` works ☺️

src/test/kotlin/ecommerce/exception/ProductExceptionTest.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.test.web.server.LocalServerPort
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+private fun Int.isEqualTo(statusCode: Int) {}
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
+class ProductExceptionTest {
```
da-nyee, line 21: Good job — you tested the error cases as well 👍

src/test/kotlin/ecommerce/exception/ProductExceptionTest.kt
```diff
@@ -0,0 +1,227 @@
...
+import org.junit.jupiter.api.Test
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.boot.test.web.server.LocalServerPort
+import org.springframework.http.HttpStatus
+import org.springframework.test.annotation.DirtiesContext
+
+private fun Int.isEqualTo(statusCode: Int) {}
+
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
```
da-nyee, line 20: This might negatively impact performance.
There's an alternative — using `@Transactional`.

da-nyee: I have one more question — does `@Transactional` work differently in production and in tests?

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
```diff
@@ -0,0 +1,161 @@
...
+    fun setUp() {
+        RestAssured.port = port
+    }
+
+    @Test
+    fun `register with valid data should return 201 Created`() {
+        val registerRequest =
+            RegisterRequest("valid2@example.com", "SecureP@ss1", "User1")
+        val response =
+            RestAssured.given().log().all()
```
da-nyee, line 32: Good use of `RestAssured` 👍

da-nyee: Just curious — have you had any difficulties writing E2E tests?

da-nyee (changes requested):
Hi Apoorva! 🙌
I know this one's a bit tricky, but you've been doing great so far — keep it up 💪

I've gone through the code and left a few comments — please take a look.
If anything's unclear, feel free to reach out — I'd be happy to help ☺️

src/main/kotlin/ecommerce/controller/product/ProductController.kt
Apoorva-S-Natesh: Thank you! that makes the endpoints clear and short!

src/main/kotlin/ecommerce/exception/GlobalExceptionHandler.kt
Apoorva-S-Natesh: Good point—grouping them does make the code cleaner and avoids repetition. I'll update it accordingly, thank you.

src/main/kotlin/ecommerce/dto/product/UpdateProductRequest.kt
Apoorva-S-Natesh: Good catch! They do share a lot of fields—combining them into a single request model with will simplify things. I’ll refactor accordingly.

src/main/kotlin/ecommerce/service/LoginMemberArgumentResolver.kt
Apoorva-S-Natesh: Thank you for this question.
How I understood the difference between Authentication and Authorization is:
**Authentication** is focused on verifying the identity of a user. It answers the fundamental question: "Is this user who they claim to be?" 
When a user attempts to log in (e.g., /api/auth/login), they provide credentials (username/email and password). - So I use JWT to check if they are who they are. Upon successful verification, a signed JWT token is issued to the client. This token serves as proof of their authenticated identity for subsequent requests.
**Authorization** (What are you allowed to do?) Once a user's identity has been successfully authenticated, the authorization layer determines what actions and resources that specific user is permitted to access or modify.

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
Apoorva-S-Natesh: You are right! this function was added but is never used. I will remove the unused methods.

src/main/kotlin/ecommerce/security/JwtRequestFilter.kt
Apoorva-S-Natesh: Thank you for mentioning it, I will check that out. 
Please let me know if what I understood is correct:
Filters intercept requests before they reach the DispatcherServlet, making them ideal for coarse-grained tasks such as:
Authentication, Logging and auditing. Filters, can manipulate requests before they reach our controllers and outside of Spring MVC

src/main/kotlin/ecommerce/service/LoginMemberArgumentResolver.kt
Apoorva-S-Natesh: I had skipped the authorization in the first PR, but have added it now. I have problem with testing it though.

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
Apoorva-S-Natesh: Thank you for mentioning about indexes. I did not know about them but its quite interesting!

src/test/kotlin/ecommerce/exception/ProductExceptionTest.kt
Apoorva-S-Natesh: I have removed DirtiesContext and used `@Sql` and `@Transactional`

This was spoken today in lecture that in tests, the entire test method runs in one transaction
and Changes made during a test don't persist to the database.
During production Transactions commit when the method completes successfully and all changes are permanently saved to the database also only rolls back on unchecked exceptions.

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
Apoorva-S-Natesh: I had a lot of difficulties with the tests. Especially when I tried to use MockitoBean and the E2E test. Also had issues with persistence, when all test classes were run together, a lot of tests failed! I had tough time fixing them. Please let me know how to handle such cases. I would love to hear your approach and thoughts on it.

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
Apoorva-S-Natesh: Oh! I had misunderstood, fixed the query

src/main/kotlin/ecommerce/security/AdminAuthInterceptor.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.security.core.Authentication
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.security.core.userdetails.User
+import org.springframework.stereotype.Component
+import org.springframework.web.servlet.HandlerInterceptor
+
+@Component
+class AdminAuthInterceptor(
+    private val memberService: MemberService,
+) : HandlerInterceptor {
```
da-nyee, line 18: You made it 👍

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,114 @@
...
+    fun findRecentlyActiveMemberIds(since: LocalDateTime): List<UUID> {
+        val sql =
+            """
+            SELECT m.id
+            FROM members m
+            INNER JOIN cart_items ci ON m.id = ci.member_id
+            WHERE ci.created_at >= ?
+            GROUP BY m.id
+            ORDER BY MAX(ci.created_at) DESC
+            """.trimIndent()
```
da-nyee: You did a good job, but there's one more thing that needs to be fixed.

According to the requirement:

image

<br/>
<br/>

The current query only returns `member_id`, but it should also include the member's `name` and `email` 👀

src/test/kotlin/ecommerce/exception/ProductExceptionTest.kt
da-nyee: You know the concept very well 💯

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
da-nyee: Could you explain more about the issue you faced when testing E2E with `@MockBean`?
It would help me understand the situation more clearly.

As for the persistence issue, I think it's important to ensure that each test is independent — for example, by using `@Transactional` or cleaning up data after each test.

da-nyee (changes requested):
Hi Apoorva! 🙌

Thanks for applying the feedback — I've gone through all the changes.
However, there's one thing that doesn't meet a requirement, so I'll keep this PR in `changes requested` status for now.
Once that's fixed, I'll go ahead and merge it.

If anything's unclear, feel free to let me know 🙇‍♀️

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
Apoorva-S-Natesh: You are right, the requirement asks for member_id, email and name. In this case, the repository returns the list of member IDs, and the service returns the ActiveMemberResponse DTO with the name and email. Would you suggest doing it in a different way?
```
AdminService.kt

    fun getRecentlyActiveMembers(): List<ActiveMemberResponse> {
        val sevenDaysAgo = LocalDateTime.now().minusDays(7)
        val activeMemberIds = cartItemRepository.findRecentlyActiveMemberIds(sevenDaysAgo)
        val activeMembers = memberRepository.findByIds(activeMemberIds)
        return activeMembers.map { member ->
            ActiveMemberResponse(
                memberId = member.id,
                memberName = member.name,
                memberEmail = member.email,
            )
        }
    } 
```

Apoorva-S-Natesh:
The test was designed to mock the AuthService (which is a dependency of AuthController) and verify that the AuthController correctly handles HTTP requests and calls the mocked service.
The Problem: The @WebMvcTest annotation, attempted to configure parts of the application that it thinks are necessary, including the Spring Security filter chain. The JwtRequestFilter is part of this chain, and its constructor requires a JwtTokenProvider bean.
Since @WebMvcTest is a "slice" test, it did not load all @Component or @Service beans from the entire application. The JwtTokenProvider, while annotated as a @Component, was not available in this limited test context. So I got: UnsatisfiedDependencyException

Another problem I faced with the tests was even though tests were annotated as @Transactional, when all the tests were run together a lot of them failed. Then I was introduced to `@Sql` with scripts to clean the table before each test.
`@Sql(scripts = ["/sql/cleanup-members.sql"], executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD)`

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
Apoorva-S-Natesh: I've refactored the findRecentlyActiveMembers method to use a single, optimized SQL query. This change moves the logic for defining "active members" directly into the repository layer.

After further thought I get what you were trying to convey, my understanding now is that encapsulating business logic that is specific to data retrieval, like this, within the repository is a good practice. It reduces the number of database calls and keeps the service layer focused on orchestrating higher-level operations.
Please give your feedback.

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
```diff
@@ -0,0 +1,127 @@
...
+            FROM 
+                members m
+            INNER JOIN 
+                cart_items ci ON m.id = ci.member_id
+            WHERE 
+                ci.created_at >= ?
+            GROUP BY 
+                m.id
+            ORDER BY 
+                MAX(ci.created_at) DESC
```
da-nyee, line 117: 👍

src/main/kotlin/ecommerce/repository/CartItemRepository.kt
da-nyee: Awesome 💯
You refactored the code exactly as intended — that's what the requirement was about.

src/test/kotlin/ecommerce/controller/AuthControllerTest.kt
da-nyee: > The test was designed to mock the AuthService (which is a dependency of AuthController) and verify that the AuthController correctly handles HTTP requests and calls the mocked service.
The Problem: The `@WebMvcTest` annotation, attempted to configure parts of the application that it thinks are necessary, including the Spring Security filter chain. The JwtRequestFilter is part of this chain, and its constructor requires a JwtTokenProvider bean.
Since `@WebMvcTest` is a "slice" test, it did not load all `@Component` or `@Service` beans from the entire application. The JwtTokenProvider, while annotated as a `@Component`, was not available in this limited test context. So I got: UnsatisfiedDependencyException

To solve this, you could either mock the `JwtTokenProvider` with `@MockBean` or provide a minimal test-specific implementation via `@TestConfiguration`.

da-nyee: > Another problem I faced with the tests was even though tests were annotated as `@Transactional`, when all the tests were run together a lot of them failed. Then I was introduced to `@Sql` with scripts to clean the table before each test.
`@Sql(scripts = ["/sql/cleanup-members.sql"], executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD)`

Okay, it seems that works well to reset the data 👀

da-nyee (approved):
Hi Apoorva! 🙌

Everything's looking good to wrap up the mission.
You've been doing great, even though it's been a tough challenge.

I really hope the feedback was helpful and supported your growth.
If you have any further questions, feel free to reach out — I'd be happy to help ☺️

PR #33, opened 2025-07-28
[spring-ecommerce-product - Step-2] Submit Mariia's mission

mariyakulikova:
This pull request implements Step 2 of the eCommerce project. It includes validation and error handling using custom exceptions with the `@ResponseStatus` annotation. User registration and login for existing users are now available. The product cart feature is implemented, allowing users to add items to their cart. It also includes a statistics page showing the most popular products and the most active users.

Authorization:
- `POST /api/members/login`: login with email and password
- `POST /api/members/register`: register new user with email and password

Cart (requires logined user):
- `POST /cart` add to cart product
- `GET /cart` return product in cart
- `DELETE /cart/{productId}` remove product from cart

Product: 
- `POST /api/products/` add product to db
- `GET /api/products` return products
- `PUT /api/products/{productId}` update product
- `DELTE /api/products/{productId}` delete product
- `GET /products` return product view

Statistic (access for users with role `ADMIN`):
- `GET /stats/top-products` return list of top 5 added products
- `GET /stats/active-members` return list of active members 

This is my first Spring project, and I had to work quickly to meet the deadline. Because of that, I’m not confident about the overall structure of the code or the way I organized layers. Also, writing tests was especially difficult. I know this part of the project needs improvement.

I would be very grateful for any feedback.


build.gradle.kts
```diff
@@ -20,11 +20,17 @@ repositories {
...
+
     implementation("org.springframework.boot:spring-boot-starter-jdbc")
     implementation("org.springframework.boot:spring-boot-starter-thymeleaf")
     implementation("org.springframework.boot:spring-boot-starter-web")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
     implementation("org.jetbrains.kotlin:kotlin-reflect")
+    implementation("org.springframework.boot:spring-boot-starter-validation")
+    implementation("io.jsonwebtoken:jjwt-api:0.12.6")
+    runtimeOnly("io.jsonwebtoken:jjwt-impl:0.12.6")
+    runtimeOnly("io.jsonwebtoken:jjwt-jackson:0.12.6")
```
pkeugine, line 33: What are the differences between `implementation` and `runtimeOnly`?
I'm curious about the reason why you used `implementation` for `jjwt-api` and `runtimeOnly` for others.

src/main/kotlin/ecommerce/auth/AdminOnlyInterceptor.kt
```diff
@@ -0,0 +1,58 @@
...
+    ): Boolean {
+        if (handler !is HandlerMethod) return true
+
+        val method = handler.method
+        val hasAdminOnly = method.getAnnotation(AdminOnly::class.java) != null
+        if (!hasAdminOnly) return true
+
+        val token =
+            request.getHeader("Authorization")?.removePrefix("Bearer ")
+                ?: throw AuthorizationException("Authorization header is missing")
```
pkeugine: When it comes to response body of `UNAUTHORIZED` response, I prefer not giving it a body with error information, unless we really need it.
This is because we don't want people to find out what's wrong when they fail to breach into our system. I'm curious about your opinion on this :)

src/main/kotlin/ecommerce/exception/AuthorizationException.kt
```diff
@@ -0,0 +1,7 @@
+package ecommerce.exception
+
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.ResponseStatus
+
+@ResponseStatus(HttpStatus.UNAUTHORIZED)
+class AuthorizationException(message: String) : RuntimeException(message)
```
pkeugine: Nice use of `@ResponseStatus` 👍 
Are there any other way of handling exceptions in Spring Boot? It would be nice for you to find some and compare them. Then, let's choose the one that fits best for our situation. It's also fine to keep this way after learning other options. :)

src/main/kotlin/ecommerce/repository/CartRepository.kt
```diff
@@ -0,0 +1,29 @@
+package ecommerce.repository
+
+import ecommerce.dto.CartItem
+
+interface CartRepository {
```
pkeugine, line 5: It's interesting to see you've extracted interfaces for repositories.
I'm curious, what was the motive behind this?

In my opinion, since we only have one implementation for each interface,
I would extract an interface when we bring in other implementations of it.

src/main/kotlin/ecommerce/repository/JdbcCartStatRepository.kt
```diff
@@ -0,0 +1,59 @@
...
+                p.name AS product_name,
+                COUNT(*) AS added_count,
+                MAX(ci.added_at) AS last_added_at
+            FROM cart_items ci
+            JOIN products p ON ci.product_id = p.id
+            WHERE ci.added_at >= DATEADD('DAY', -30, CURRENT_TIMESTAMP)
+            GROUP BY ci.product_id, p.name
+            ORDER BY added_count DESC, last_added_at DESC
+            LIMIT 5
+
```
pkeugine:
```suggestion
            SELECT 
                ci.product_id,
                p.name AS product_name,
                COUNT(*) AS added_count,
                MAX(ci.added_at) AS last_added_at
            FROM cart_items ci
            JOIN products p ON ci.product_id = p.id
            WHERE ci.added_at >= DATEADD('DAY', -30, CURRENT_TIMESTAMP)
            GROUP BY ci.product_id, p.name
            ORDER BY added_count DESC, last_added_at DESC
            LIMIT 5
```

Let's keep the indentation consistent and remove unnecessary line.
Please check other queries in the project and apply this where it's needed 👍

src/main/kotlin/ecommerce/dto/Product.kt
```diff
@@ -0,0 +1,22 @@
...
+        regexp = """^[\w\s()\[\]+\-&/_]*$""",
+        message = Constants.ERR_NAME_REGEX,
+    )
+    @field:Size(max = 15, message = Constants.ERR_NAME_SIZE)
+    var name: String,
+    @field:Positive(message = Constants.ERR_PRICE_POSITIVE)
+    var price: Double,
+    @field:Pattern(regexp = """^(http://|https://).+""", message = Constants.ERR_URL_REGEX)
+    var imageUrl: String,
+)
```
pkeugine: I think the current dto classes have too much responsibilities. It contains validation, are being used as request / response bodies of controllers, acts as domains, and some of them are used as entities as well. Why don't we divide the responsibilities to multiple classes so that entities (or domains) & request / response bodies are treated separately?

It would be fun for you to think about why I left this comment as well :)

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+import java.nio.charset.StandardCharsets
+import java.util.Date
+import javax.crypto.SecretKey
+
+@Component
+class JwtTokenProvider(
+    @Value("\${security.jwt.token.secret-key}")
+    secret: String,
+    @Value("\${security.jwt.token.expire-length}")
+    private val validityInMs: Long,
```
pkeugine, line 16: Using `@Value` to bring the value from `application.properties` 👍 
Why do you think keeping these values in `application.properties` is better than keeping them in application logic?

src/main/kotlin/ecommerce/controller/ProductController.kt
```diff
@@ -15,56 +16,40 @@ import org.springframework.web.bind.annotation.RestController
...
     @PutMapping("/{id}")
     fun updateProduct(
-        @RequestBody newProduct: Product,
+        @Valid @RequestBody newProduct: Product,
         @PathVariable id: Long,
     ): ResponseEntity<Unit> {
-        val updated = productRepository.update(id, newProduct)
-        if (!updated) {
-            throw RuntimeException("Product not found")
-        }
```
pkeugine, line 42: I'm curious about the reason why you removed not-found-validation when updating 😮

src/test/kotlin/ecommerce/auth/JwtTokenProviderTest.kt
```diff
@@ -0,0 +1,27 @@
+package ecommerce.auth
+
+import org.assertj.core.api.Assertions
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.test.context.SpringBootTest
+
+@SpringBootTest(properties = ["spring.sql.init.mode=never"])
```
pkeugine: I'm actually curious about [your answer to this question](https://github.com/herotechcourse/spring-ecommerce-product/pull/2#discussion_r2219339020) :)

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+        `when`(repository.create(product)).thenReturn(42L)
+
+        val response: ResponseEntity<Unit> = controller.createProduct(product)
+
+        verify(service).validateUniqueName(product)
+        verify(repository).create(product)
+
+        assertEquals(HttpStatus.CREATED, response.statusCode)
+        assertEquals(URI.create("/api/products/42"), response.headers.location)
+    }
```
pkeugine: I see you've utilized Mockito 😮 
How did it feel like using it?

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
```diff
@@ -0,0 +1,66 @@
...
+            listOf(
+                Product(id = 1, name = "Apple", price = 1.0, imageUrl = "url"),
+                Product(id = 2, name = "Banana", price = 2.0, imageUrl = "url"),
+            )
+        `when`(repository.getAll()).thenReturn(products)
+
+        val response = controller.readProducts()
+
+        assertEquals(HttpStatus.OK, response.statusCode)
+        assertEquals(products, response.body)
```
pkeugine, line 45: When do you use `assertJ`, and when do you use other libraries?
I ask this question because I personally try to stick with one as much as I can 😄

src/test/kotlin/ecommerce/auth/JwtTokenProviderTest.kt
```diff
@@ -0,0 +1,27 @@
...
+class JwtTokenProviderTest {
+    @Autowired
```
pkeugine: I can see that in this project, there are unit tests, mocked unit tests, and integrated tests (some call it E2E tests, but I like to call `@SpringBootTest` based tests integrated tests). I remember [you even utilized `@JdbcTest` in previous step, performing a sliced test.](https://github.com/herotechcourse/spring-ecommerce-product/pull/2#discussion_r2216564115)

Now that you've experienced different kinds of tests, which type of tests do you think are very effective for you? Would you make as much tests as possible when you manage your project? Why or why not? :)

src/test/kotlin/ecommerce/service/ProductServiceTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    @Test
+    fun `should pass silently when product name is unique`() {
+        val product = Product(name = "UniqueName", price = 10.0, imageUrl = "http://image.jpg")
+
+        every { repository.existsByName(product) } returns false
+
+        service.validateUniqueName(product) // no exception expected
+    }
```
pkeugine:
```suggestion
    @Test
    fun `should pass silently when product name is unique`() {
        val product = Product(name = "UniqueName", price = 10.0, imageUrl = "http://image.jpg")

        every { repository.existsByName(product) } returns false

        assertThatCode {
            service.validateUniqueName(product)
        }.doesNotThrowAnyException()
    }
```

It would be even better for us to assert instead of comment telling us!
Remember, tests should always contain some form of assertions.
Let's check other tests as well and see if we can apply this 😄

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
```diff
@@ -0,0 +1,57 @@
...
+                Jwts.parser()
+                    .verifyWith(secretKey)
+                    .build()
+                    .parseSignedClaims(token)
+            !claims.payload.expiration.before(Date())
+        } catch (e: JwtException) {
+            false
+        } catch (e: IllegalArgumentException) {
+            false
+        }
```
pkeugine:
```suggestion
        return runCatching {
            val claims = Jwts.parser()
                .verifyWith(secretKey)
                .build()
                .parseSignedClaims(token)

            !claims.payload.expiration.before(Date())
        }.getOrDefault(false)
```

We can also utilize `runCatching` from Kotlin :)
Though it's not 100% same as before, I think other Exceptions thrown inside this scope could also be treated similarly to `JwtException` and `IllegalArgumentException`.

src/main/kotlin/ecommerce/auth/LoginMember.kt
```diff
@@ -0,0 +1,5 @@
+package ecommerce.auth
+
+@Target(AnnotationTarget.VALUE_PARAMETER)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class LoginMember
```
pkeugine, line 5: Nice annotation! I still remember creating my very first annotation as a developer :)
What are the features of `@Target` annotation and `@Retention` annotation?
What are the other values that could go inside in those parenthesis?

pkeugine (changes requested):
Hello Mariia, although you were in a hurry, your project looks very solid 👊 

I left some questions, suggestions, and some chats(?) that ended up being questions 😄 
When you're done going through all of them, feel free to request me for another review.

If you have any questions, please DM me or leave a comment anytime.
Happy Hacking! 👊 👊 👊

src/main/kotlin/ecommerce/auth/AdminOnlyInterceptor.kt
mariyakulikova: Thanks for pointing this out. I honestly didn’t think about it from that security perspective.
I’ve removed body from response.

src/main/kotlin/ecommerce/repository/CartRepository.kt
mariyakulikova: In my case, I decided to extract interfaces because I thought we might need to change the internal logic of the repositories later (switch from JDBC to JPA). So I wanted to isolate that detail behind an interface.
Also, I wanted to practice using interfaces in a realistic context, it felt like a good opportunity to build the habit.
Right now, my repository classes have a Jdbc prefix, but in the next mission we'll be replacing them with JPA-based implementations, so the separation should help with that transition.

src/main/kotlin/ecommerce/exception/AuthorizationException.kt
mariyakulikova: Yes, I looked into a few options:
1. `@ResponseStatus` – simple and readable.
2. `@ExceptionHandler` in a `@ControllerAdvice` – more flexible, allows custom body/logging.
3. `ResponseEntityExceptionHandler` – useful for overriding default Spring behavior.

Since our use case is quite simple, I think `@ResponseStatus` works fine for now.

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
mariyakulikova: Thanks! 🙂

I put these values in `application.properties` because:
1. It’s easier to change them later without touching the code.
2. I can use different values for environments.
3. It's safer. We shouldn’t keep secrets like token keys in the code.
4. It makes the code cleaner, only logic stays in the class.

By the way, is the JWT secret key considered sensitive information?  
Should we store it separately (like in environment variables) and avoid pushing it to GitHub?

src/test/kotlin/ecommerce/auth/JwtTokenProviderTest.kt
mariyakulikova: Thank you! 🙂

After trying different types of tests, I found that:
- `Unit tests` are great for checking logic quickly and run super fast.
- `Mocked tests` help isolate logic and avoid depending on real services.
- `Integration tests` (like `@SpringBootTest`) are the most helpful when checking if everything works together.

To be honest, I still don’t feel very confident writing tests yet. It’s something I’m still learning, and sometimes I’m not sure what exactly should be tested or how deep I should go.
But I’d try to cover the important parts of the system, especially things that are hard to test manually or easy to break.

src/main/kotlin/ecommerce/auth/LoginMember.kt
mariyakulikova: Here’s what I learned:
- `@Target` defines where the annotation can be used.  
  For example:
  - `VALUE_PARAMETER` – for function parameters
  - `CLASS`, `FIELD`, `FUNCTION`, `PROPERTY`, etc.

- `@Retention` defines how long the annotation is kept:
  - `SOURCE` – removed at compile time
  - `BINARY` – present in the .class file, but not at runtime
  - `RUNTIME` – available at runtime (needed for reflection)

Since I use this annotation to inject the logged-in member using an argument resolver, I chose `VALUE_PARAMETER` + `RUNTIME`.

build.gradle.kts
mariyakulikova: What I learnt:
- `implementation` is for libraries we need at compile time and runtime.
- `runtimeOnly` is for libraries only needed at runtime, they are not used directly in the code.

I added these lines because they were in the course materials.  
From what I saw, if we want to use the latest version of the `jjwt` library (0.12.6), we need to include all three parts:  
- `jjwt-api` (the interface),
- `jjwt-impl` (the actual implementation),
- and `jjwt-jackson` (for JSON serialization).

Honestly, it was hard to understand everything 😅  
I just followed the suggested setup from the course materials.

src/main/kotlin/ecommerce/controller/ProductController.kt
mariyakulikova: I removed the not-found check from the controller and moved the exception into the repository.  
But now I realize that it’s not the repository’s responsibility to handle this, so I moved the check to the service layer instead.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
mariyakulikova: I sort of know what it's supposed to do, but it still feels like dark magic sometimes.

src/test/kotlin/ecommerce/controller/ProductControllerTest.kt
mariyakulikova: I think I mixed them mostly out of habit. I started with assertEquals because it's simple and familiar, but I do see the value in sticking to one consistent assertion style.
I’ll try to get used to using just one library going forward. Thanks for pointing it out!

src/main/kotlin/ecommerce/dto/Product.kt
mariyakulikova: On the `ecommerce-order` mission, this issue actually became very clear. I realized that separating DTOs from domain models is essential.
Having a single class handle validation, request/response mapping, and business logic leads to tight coupling and makes testing or refactoring harder.

src/test/kotlin/ecommerce/service/ProductServiceTest.kt
mariyakulikova: Thanks for the suggestion. I updated the test to include an explicit assertion now.

build.gradle.kts
pkeugine: I think that's the process of learning :)
Nice job explaining the difference 👍

src/main/kotlin/ecommerce/auth/JwtTokenProvider.kt
pkeugine: Yes, I think it would be great to store them separately, but since this is our first time trying it out, I understand it getting exposed in GitHub. From now on, let's try not to reveal it to the public :)

src/main/kotlin/ecommerce/auth/LoginMember.kt
pkeugine: Nice! 💯

pkeugine (approved):
Hello Mariia, and nice work finishing this mission!
It must have been hard keeping up with this while also working on the current mission.
I respect your hard work 👍 

I see you've refactored the project and answered all the questions (including the comment on the previous PR), so I think we are ready to let this PR get merged.

I hope you do well in current mission with the things we learned here,
and continue on with the good work throughout the course!

Happy Hacking! 🔥 🔥 🔥

